PR #118, opened 2026-01-27
feat: Startup cache refresh, dark mode fixes, regex testing modal

Sp0rtsFr3ak:
## Changes

  1. **Configurable startup cache refresh** - Users can set cache to auto-refresh if older than 1/3/7
  days (default: 1 day)

  2. **Checkbox/switch visibility fix** - Components now visible in dark mode (Tailwind v4
  compatibility)

  3. **Create/Test Patterns modal** - Full-featured regex pattern builder with live stream testing,
  color-coded highlighting, and pattern suggestions

  ## Testing
  - Tested in Docker
  - All three features verified working

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -4,6 +4,7 @@ import { MainLayout } from "@/layouts/MainLayout"
 import { GenerationProvider } from "@/contexts/GenerationContext"
 import { StartupOverlay } from "@/components/StartupOverlay"
 import {
+  AI,
   Dashboard,
   Templates,
   TemplateForm,
@@ -92,6 +93,7 @@ function AppContent() {
             <Route path="event-groups/import" element={<EventGroupImport />} />
             <Route path="epg" element={<EPG />} />
             <Route path="channels" element={<Channels />} />
+            <Route path="ai" element={<AI />} />
             <Route path="settings" element={<Settings />} />
           </Route>
         </Routes>
--- a/frontend/src/layouts/MainLayout.tsx
+++ b/frontend/src/layouts/MainLayout.tsx
@@ -1,18 +1,17 @@
 import { Link, NavLink, Outlet } from "react-router-dom"
-import { Moon, Sun } from "lucide-react"
-import { useEffect, useState } from "react"
+import { Moon, Sun, Brain } from "lucide-react"
+import { useEffect, useState, useMemo } from "react"
 import { Toaster } from "sonner"
 import { useQuery } from "@tanstack/react-query"
-import { useUpdateCheckSettings, useCheckForUpdates } from "../hooks/useSettings"
+import { useUpdateCheckSettings, useCheckForUpdates, useAISettings } from "../hooks/useSettings"
 
-const NAV_ITEMS = [
+const BASE_NAV_ITEMS = [
   { to: "/", label: "Dashboard" },
   { to: "/templates", label: "Templates" },
   { to: "/teams", label: "Teams" },
   { to: "/event-groups", label: "Event Groups" },
   { to: "/epg", label: "EPG" },
   { to: "/channels", label: "Channels" },
-  { to: "/settings", label: "Settings" },
 ]
 
 async function fetchHealth(): Promise<{ status: string; version: string }> {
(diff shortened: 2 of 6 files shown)


egyptiangio:
**PR #118 Review**

Thanks for this contribution — all three features address real issues and the implementations are thoughtful. I've taken a different approach on each, but your work directly influenced my solutions.

---

**Feature 1: Startup Cache Refresh**

Your observation about frequent provider calls is valid. I opted to keep the startup refresh — most users don't restart frequently, so the ~15 seconds is acceptable and guarantees fresh data. For developers or testers who restart often, I added a `SKIP_CACHE_REFRESH=true` env var to skip it. I didn't expose this in the UI to avoid confusion for regular users.

I also increased the scheduled refresh frequency from every 7 days to every 1 day (no configurability), and kept the manual trigger in settings for when users know they need fresh data.

I won't be merging the settings UI/migration, but appreciate you flagging the issue.

---

**Feature 2: Dark Mode Checkbox/Switch**

Great catch — they were indeed invisible. The root cause was deeper: the `@theme` colors used hex format, and Tailwind v4 can't apply alpha modifiers (`/50`, `/10`) to hex. This broke 161 usages across 40+ files, not just checkboxes.

I converted all colors to `oklch()` format — one CSS change that fixed everything. Also did a standardization pass on checkbox patterns codebase-wide. Your bug report led me to a much better fix.

---

**Feature 3: Regex Test Patterns Modal**

This feature idea is excellent — regex for IPTV streams is genuinely hard, and the placeholder had been there too long. I've now shipped a full implementation inspired by your concepts:

- Bidirectional sync with the actual form fields (not a separate UI)
- Interactive text selection to generate regex with named capture groups
- Live match highlighting

Several of your ideas (Python-to-JS regex conversion, color highlighting, live testing) made it into the final version. Thank you for the inspiration.

---

Closing this PR since I've implemented alternatives for all three features. Really appreciate the contribution and the bug reports that came with it.

Sp0rtsFr3ak:
> **Re: European timezone date adjustment ([d5c86b4](https://github.com/Pharaoh-Labs/teamarr/commit/d5c86b4f2cbce0fb3018597838118f67bc11514d) - [3c9022f](https://github.com/Pharaoh-Labs/teamarr/commit/3c9022fd99c0fcd1fa5dfe3cb7b7a9e199c8b929))**
> 
> I want to understand this better before deciding. The extracted date is only used as a filter — if a stream has no date in the name (most IPTV streams), this code path is skipped entirely.
> 
> A few questions:
> 
> 1. Do you have real examples of streams with embedded dates that are failing to match? Or is this theoretical?
> 2. The -1 day heuristic seems fragile — Australian/Asian streams would get wrong adjustments, as would actual midnight US games.
> 3. This same problem exists in reverse, right? A game in London shown with US times would have the opposite issue.
> 
> If this is a real problem you're hitting, I'd prefer proper timezone conversion when a marker (CET, GMT, ET) is present, rather than assuming -1 day. If no timezone marker, we probably can't reliably adjust at all.
> 
> Would love to see some example stream names that are failing for you.

This was for an edge case I was dealing with. I had 1 group with the date and time in 24 hr format. And it was the UK timezone, no timezone markers at all. So games weren't matching because Teamarr couldn't find a game card for the date of the game. The stream names in this particular folder often showed games at 01:00 the following day, since the UK is 7 hours ahead of me. This fix tries to address that, and it's based off your local time. I've been testing and monitoring results. So far, so good. Between this fix and the PPV fix, I have seen about 100 more streams getting matched. I haven't had any mismatches yet, either.

egyptiangio:
The **Stream Timezone Awareness** feature is now available in `dev`!

## What's New

**Auto-detection** - Timezone markers in stream names (ET, PT, CT, GMT, etc.) are now automatically extracted and used for date matching. The regex helper UI also captures these when you select a time.

**Per-group timezone setting** - New "Stream Timezone" card in group settings lets you specify the provider's timezone when markers aren't included in stream names.

**Bulk support** - Set timezone during bulk import or bulk edit.

## How It Works

1. If stream includes timezone marker (e.g., "9:00 PM ET") → auto-detected
2. If no marker but group has timezone set → uses group setting
3. Otherwise → falls back to your local timezone

This solves the date mismatch problem when your provider uses a different timezone than your local time (e.g., UK user seeing US Eastern times would have midnight crossover issues).

**UI location:** Edit any Event Group → "Stream Timezone" card (after Basic Settings)

PR #120, opened 2026-01-29
Regular tv events

nreeb:
I implemented the ability to select groups that contain regular tv channels such as TSN, SportsNet, ESPN etc., then the programmes on those channels throughout the day are created into channels on their own in a custom m3u that is sourced in teamarr.

New Settings tab - Toggles on the feature, sets a default epg file for the groups, sets a lookback and look forward values
New page for Regular TV - This is where you import the groups that have regular tv channels in it. Similar to what you do in event groups. Once you select the groups you can generate the custom m3u to assess your output (this will happen when epg is generated automatically so running it here manually is not required to make it work)
EventGroups - When you import groups you will now see the custom group for regular tv to include them

When the epg is generated it regenerates a fresh m3u and if the regular tv group is created they are now channels used in the regular processing.

The custom m3u is created by using regex to filter out events

There are a ton of "regular tv" channels that have local feeds or sporting events that may not be on the event specific groups. Especially local feeds

### DIFF
--- a/teamarr/consumers/event_group_processor.py
+++ b/teamarr/consumers/event_group_processor.py
@@ -2288,15 +2288,10 @@ class EventGroupProcessor:
         """
         from teamarr.consumers.lifecycle import StreamProcessResult
 
-        # Determine client to use - skip Dispatcharr for Regular_TV
-        client = self._dispatcharr_client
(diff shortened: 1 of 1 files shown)


nreeb:
No problem.

Thanks. I have decided It might be best to create a separate m3u generator
that can be hosted and I can add to Temarr as a team specific m3u.

PR #123, opened 2026-01-30
Fix/european timezone date matching

Sp0rtsFr3ak:


egyptiangio:
The **Stream Timezone Awareness** feature is now available in `dev`!

## What's New

**Auto-detection** - Timezone markers in stream names (ET, PT, CT, GMT, etc.) are now automatically extracted and used for date matching. The regex helper UI also captures these when you select a time.

**Per-group timezone setting** - New "Stream Timezone" card in group settings lets you specify the provider's timezone when markers aren't included in stream names.

**Bulk support** - Set timezone during bulk import or bulk edit.

## How It Works

1. If stream includes timezone marker (e.g., "9:00 PM ET") → auto-detected
2. If no marker but group has timezone set → uses group setting
3. Otherwise → falls back to your local timezone

This solves the date mismatch problem when your provider uses a different timezone than your local time (e.g., UK user seeing US Eastern times would have midnight crossover issues).

**UI location:** Edit any Event Group → "Stream Timezone" card (after Basic Settings)

Issue #22, opened 2025-12-07
Fallback to Dispatcharr dummy EPG for unmatched streams


egyptiangio:
## Summary

Allow unmatched streams (that aren't excluded by filters) to still get channels created, using Dispatcharr's native dummy EPG as a fallback instead of ESPN event data.

## Use Case

- Streams that don't match ESPN events but user still wants in their EPG
- Events from leagues/sports not covered by ESPN
- Streams with non-standard naming that can't be parsed

Currently these show as "unmatched" and get no channel. User may want them visible anyway with basic placeholder EPG.

## Proposed Implementation

Per-group setting:

1. Add toggle: "Create channels for unmatched streams"
2. When enabled, user selects which Dispatcharr dummy EPG source to use
3. All unmatched streams in that group (not excluded by filters) get:
   - Channel created
   - Associated with selected dummy EPG
4. All-or-none per group (no per-stream selection)

## Considerations

- Channel naming: Use stream name directly since no event data available
- Channel numbering: Would need to fit into existing numbering scheme
- Cleanup: These channels would follow same lifecycle rules

## Priority

Low - tracking for future implementation

egyptiangio:
Related to #13 - both requesting fallback channel creation for unmatched streams.

egyptiangio:
Reconsidering scope on this one.

Current focus is capturing >90% of streams by:
- Improving the matching engine
- Adding more supported sports/leagues
- Better pattern recognition

Passthrough to Dispatcharr autocreate/dummy creates a layer of complexity and maintenance burden I'm not sure I want to take on. The integration surface area with Dispatcharr is already significant.

Will reevaluate once V2 comes out of alpha and we have a better picture of real-world match rates.

egyptiangio:
Closing as out of scope.

Teamarr's core objective is accurate event-to-stream matching with authoritative EPG data. Building passthrough mechanisms for unmatched streams dilutes this focus and introduces complexity without improving match coverage.

For streams that can't be matched, the better path is improving the matching algorithm and/or expanding sport/league coverage rather than creating fallback channels with placeholder data.

Unmatched streams are still accessible directly in Dispatcharr if needed.

PR #125, opened 2026-01-30
feat(event-groups): add unmatched stream channel handling

lpukatch:
Full disclosure I used gemini to assist with these changes

Based on discussion in #22 

This adds db fields to support creation of unmatched channels and dummy epg for items that are not matched. 

Added a process_unmatched teams method that creates / removes channels after the matched channels are added, then adds the dummy epg.

EventGroupForm and EventGroupInput pull dummy EPGs and present in the form when adding or editing an event group


lpukatch:
Going to redo these changes on a new branch after the major refactoring

PR #130, opened 2026-02-05
Date time format platform agnostic

nreeb:
The datetime formats are not agnostic. Works fine on linux machines, I have been testing on a windows machine and they will not work and throw exceptions. I have changed the formats to determine the platform and use the appropriate format that will work on windows or linux


egyptiangio:
Merged in v2.1.2 - thanks!

PR #142, opened 2026-02-08
Backup System

Sp0rtsFr3ak:


egyptiangio:
Thanks for the contribution! Merged to dev.

I'll be making some follow-up fixes for integration with the existing codebase:
- **Migration cleanup** — the v52/v53 migration blocks got overwritten, restoring those and adding v54 cleanly
- **Scheduler integration** — folding `BackupScheduler` into the existing `CronScheduler` (same pattern as channel reset)
- **Path validation** — adding traversal checks on filename endpoints and backup path setting
- **Safe DB copy** — switching to `sqlite3.backup()` instead of `shutil.copy2()` on live DB

The core `BackupService`, UI, and hooks are solid — nice work on the protection/rotation system.

PR #169, opened 2026-03-08
Edge case with detect_sport_hint() returning list instead of string

thepagan:
fix(groups): handle list return from detect_sport_hint in builtin stream filter

detect_sport_hint() can return either a string or a list depending on regex matches (notably with streams with custom regex titles). The previous implementation assumed a string and attempted membership checks against UNSUPPORTED_SPORTS, causing:

TypeError: unhashable type: 'list'

The builtin filter logic now safely handles both string and list values and checks each sport hint individually.

### DIFF
--- a/teamarr/api/routes/groups.py
+++ b/teamarr/api/routes/groups.py
@@ -1702,8 +1702,15 @@ def get_raw_streams(group_id: int):
         if is_placeholder(name):
             return "placeholder"
         sport = detect_sport_hint(name)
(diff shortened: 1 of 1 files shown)


egyptiangio:
Good catch! This was a bug we introduced in the multi-sport hint work — `detect_sport_hint()` can now return a list (e.g., `["Soccer", "Football"]` for ambiguous terms like "English Football League") but this caller still assumed a string.

Took your fix and simplified it slightly in `7aa517e` on dev. Thanks for the contribution!

PR #182, opened 2026-03-26
Feature: Add support for managing Headendarr also

Josh5:
This PR adds support for Headendarr (https://github.com/Headendarr/Headendarr) as an additional Teamarr integration target alongside the existing Dispatcharr integration.

Headendarr is an open-source IPTV/channel management application that can manage playlists, channels, guide sources, and playback source failover. This change allows Teamarr to provision and update Teamarr-generated XMLTV data into Headendarr, import Headendarr playlists for event matching, and create/update Headendarr-backed channels using Teamarr’s existing sports guide generation and scheduling workflows.

Note:

These changes do not affect existing Dispatcharr users keep the current flow unchanged.
It does not change existing Teamarr template, scheduling, and generation behaviour remains intact unless Headendarr is explicitly enabled.

Along side the Headendarr support changes, I also added support for adding xmltv_ns tags to the EPG. Eg. `<episode-num system="xmltv_ns">...</episode-num>`.

Some DVR clients airings differently if this tag exists. In Plex specifically, this can change sports recordings from behaving more like standalone movie-style items to behaving more like TV-series episodes.

So instead of a recording existing like:
- `/data/media/plex/dvr/sports/Basketball League Team 1 at Team 2/Basketball League Team 1 at Team 2.ts`

it would look like:

- `/data/media/plex/dvr/tv/Basketball League/Season 2026/Basketball League - S2026E871 - Team 1 at Team 2.ts`

I've also updated the docs to present both Dispatcharr and Headendarr as supported integrations and adds Headendarr-specific setup and architecture documentation.


egyptiangio:
## PR Review — Headendarr Integration

> **Disclosure:** This review was conducted with the assistance of Claude (AI). All findings were verified against the actual patch, but please flag anything that looks incorrect.

Hey @Josh5 — thanks for this PR. The scope is impressive and the architecture closely mirrors our existing Dispatcharr integration, which made it easy to follow. The source abstraction in the import flow is clean, and the EPG provisioning workflow is well thought out. There's a solid foundation here.

That said, there are several issues we'd need addressed before merging. I've grouped them by severity.

---

### Must Fix

**1. PR targets `main` — should target `dev`**
All development goes through `dev` per our workflow. Please retarget the PR to `dev`.

**2. Event channels will have no logos via Headendarr**
In `lifecycle/__init__.py`, the Headendarr path sets `channel_manager` and `epg_manager` but leaves `logo_manager = None`. The lifecycle service's `_create_channel()` relies on `LogoManager.upload()` to get a `logo_id`, then passes `logo_id=` to `create_channel()`. But Headendarr's `create_channel()` explicitly deletes `logo_id` (`del channel_group_id, logo_id, ...`) and expects `logo_url` instead. Result: every event channel created via the lifecycle path will have no logo.

The team channel path in `_sync_headendarr_team_channels` correctly passes `logo_url` — so this is just the lifecycle event-channel path that's broken.

**3. `create_channel` can return `success=True` with `channel=None`**
After POSTing to `/tic-api/channels/new`, the code looks up the new channel via `find_by_tvg_id()` / `find_channel_by_name()` — but these search the **pre-creation cache**. If the lookup misses (stale cache, API lag), it returns `OperationResult(success=True)` with no channel data. The lifecycle service then stores `dispatcharr_channel_id=None` in the DB, permanently orphaning the channel.

Fix: either parse the channel from the POST response body, or force a cache refresh before the lookup.

---

### Should Fix

**4. No thread lock on singleton factory**
`get_factory()` uses check-then-set on `global _factory` without a lock. Our Dispatcharr factory uses `threading.Lock()` for this. Concurrent startup requests can create duplicate factory instances.

**5. `close_headendarr()` doesn't reset `_factory = None`**
After close, `_factory` remains set with a dead connection. Our `close_dispatcharr()` resets it to `None`. Calling `get_factory()` after close will return the stale instance.

**6. Layer violation — consumer imports from API routes**
`generation.py` imports `HEADENDARR_TEAMARR_EPG_NAME`, `_build_teamarr_xmltv_url` from `teamarr.api.routes.settings.headendarr`. Consumers should never import from routes. These should live in `teamarr/headendarr/` (e.g., a `constants.py`). The EPG name constant is also defined separately in `managers/channels.py` — 3 places total.

**7. Source-accounts fallback is Dispatcharr-first — dual-source not possible**
`list_source_accounts()` tries Dispatcharr first; if it returns anything (even 0 accounts), Headendarr playlists are never shown. A user with both integrations configured can only see Dispatcharr sources in the import sidebar. Same for `list_source_groups()` and `list_source_group_streams()`. Not sure if dual-source is a goal, but worth noting.

**8. `preview_group` fallback broken for Headendarr-only users**
```python
conn = factory.get_connection() if factory else get_headendarr_connection(get_db)
```
`factory` is always truthy (created at startup even when Dispatcharr is unconfigured), so the Headendarr fallback never fires. Needs `if factory and factory.is_configured`.

**9. Status endpoint polls the real Headendarr server every 30s**
`useHeadendarrStatus` has `refetchInterval: 30000`, and `/headendarr/status` calls `test_connection()` which makes a real HTTP request each time. Every open browser tab hits the Headendarr server every 30s. Our Dispatcharr status caches connection state. Either cache it server-side or gate the query on `activeTab === "headendarr"`.

**10. `_find_headendarr_team_stream_ids` return type annotation is wrong**
Signature says `-> list[int]` but returns stream objects (callers use `getattr(stream, "id")`). Should be `-> list[HeadendarrStream]` or similar.

---

### Minor / Nice to Have

- **`episode_num` feature is unrelated to Headendarr** — Patch 8 adds XMLTV `xmltv_ns` episode numbering for DVR clients. Great feature, but it should be its own PR for clean bisect/revert. Happy to merge that separately.
- **Developer notes in the UI** — The "Current Scope" card mentions "lifecycle service" and "CSO" which are internal terms. Should be user-facing language or removed.
- **Two Save buttons** on the Headendarr tab (Connection card and EPG Provisioning card) both save the same form state — confusing UX. One Save at the bottom would be cleaner.
- **`HeadendarrChannel` type** in `types.py` is defined but never used (dead code — `HeadendarrLifecycleChannel` is used instead).
- **`TokenManager`** has no thread safety (our Dispatcharr version uses `threading.Lock()`) and no token expiry tracking.
- **`_stable_group_id` uses CRC32** — 32-bit collision space. Fine for typical group counts but worth a comment.
- **`_get_teamarr_epg_id`** makes an API call every invocation — should cache since the EPG source ID won't change mid-session.
- **Headendarr tab always visible** even when not configured — adds clutter for most users. Not blocking, but could gate on `headendarr.enabled`.

---

### Schema

The schema changes are clean — 5 new nullable columns on `settings` with safe defaults, reconciliation handles upgrades. Version bump to 72 is fine once retargeted to `dev` (we're at 71 on both branches, no conflict today).

---

### Summary

| Severity | Count |
|----------|-------|
| Must fix | 3 |
| Should fix | 7 |
| Minor | 8+ |

The architecture is sound and follows our patterns well. The main gaps are around the lifecycle integration (logos, orphaned channels) and some thread safety / fallback logic issues. Once the must-fix items are addressed and the PR is retargeted to `dev`, we can re-review and get this merged.

Let us know if you'd prefer to address these yourself or if you'd like us to take it from here (squash + fix, with full contributor credit — same approach we used for PR #168). Either way works for us.

Josh5:
Nice. Leave that with me over the weekend and I'll get it all sorted.

Josh5:
I've addressed the review items and implemented the requested fixes. Most significantly, I have completed the dual-source architecture to allow Dispatcharr and Headendarr to coexist fully.

Dual-Source Support (Review Item 7)

- Schema & Data Model: Added a persisted source_type field to Event Groups (Schema v73). This allows groups with the same name/account IDs to coexist across different platforms.
- API & UI: The import flow now correctly lists accounts/playlists from both platforms simultaneously.
- Background Processing: Updated run_full_generation and all enforcement tasks (Keyword Ordering, Orphan Cleanup, Scheduled Deletions) to iterate through all active platforms. Processing is now driven by the group’s persisted source type rather than global configuration state.

Core Fixes:

- Lifecycle Logos (2): Fixed the lifecycle path to correctly pass logo_url to Headendarr during channel creation and updates.
- Orphan Prevention (3): create_channel now forces a cache refresh after creation and returns failure if the record cannot be resolved, preventing None ID orphans.
- Thread Safety & Lifecycle (4, 5): Added a singleton lock to the Headendarr factory and ensured the factory resets to None on close(). Added locking and expiry tracking to the TokenManager.
- Layer Violations (6): Moved EPG constants and URL helpers from API routes to teamarr/headendarr/constants.py.
- Fallback Logic (8): Fixed the preview/raw stream fallbacks to check factory.is_configured so they correctly trigger for Headendarr-only users.
- Optimizations (9): Gated Headendarr status polling to the settings tab and implemented caching for the Teamarr EPG source ID lookup.
- Cleanup: Corrected the \_find_headendarr_team_stream_ids return annotation (10), removed the unused HeadendarrChannel type, and removed developer-facing "Current Scope" UI elements and duplicate save buttons.

Regarding episode_num XMLTV Support
The episode_num support is included because it is essential for the Headendarr ecosystem, even if the Headendarr
API itself doesn't strictly require it.

Headendarr provisions Teamarr’s XMLTV feed into downstream DVR clients like Plex. Without `<episode-num system="xmltv_ns">`, these clients often misclassify or collapse generated sports events that share similar titles or occur on the same day. By providing deterministic episode metadata based on the airing date and ordinal, we ensure the guide behaves correctly in the DVR environments Headendarr serves. While this logic lives in the generic XMLTV generator, it is a requirement for a "production-ready" Headendarr integration.

egyptiangio:
Hey @Josh5 — checking in. `dev` has moved substantially since March (Jellyfin, Channels DVR, schema reconciliation, v73 MiLB rename, audit refactors, Apr 2026 code-health pass), and GitHub reports merge state as unknown, so this needs a rebase against current `origin/dev` before it can be reviewed.

If you're still interested in seeing Headendarr support merged and maintaining it going forward, please rebase against current `origin/dev` and we'll do a proper review pass. The `xmltv_ns` change in particular looks small and useful enough to potentially split out independently if that helps. Let us know.

egyptiangio:
Hey @Josh5 — really sorry for the delay here. You put a lot of work into this and I let it sit too long. Other priorities kept pulling focus and I never got back to it properly.

If Headendarr integration is still something you're interested in maintaining, I'd love to revisit it. A fresh PR rebased against current `dev` (we're at schema v73 now) with any outstanding review items addressed would be the path forward. Happy to prioritize the review if you pick it up again.

Issue #189, opened 2026-04-03
[V2 Bug]: event group month and day fields are not saved in event groups


broken-droid:
### Version

latest

### What happened?

any custom regex entered into the month or day fields in an event group does not save when you click update group.

### Expected behavior

Save it.

### Steps to reproduce

1. enter regex into month/day field
2. click update group

### Affected Area

UI/Frontend

### Logs

```shell

```

### Screenshots

before saving:
<img width="1317" height="275" alt="Image" src="https://github.com/user-attachments/assets/d56fda94-3911-401f-8e65-1388bc353abb" />

after saving:

<img width="1013" height="240" alt="Image" src="https://github.com/user-attachments/assets/8a057c98-d71c-46cf-9af6-7c82337e6754" />

egyptiangio:
Fixed via PR #191 — cherry-picked onto `dev`. Will be in the next release.

PR #191, opened 2026-04-03
add custom_regex_month/day related fields to the useEffect setFormData

broken-droid:
fixes the state of these not showing correctly in the UI

#189


egyptiangio:
Thanks for the fix! Cherry-picked onto `dev` (`4e496e6`). This will land on `main` with the next release. Closing in favor of the cherry-pick.

PR #196, opened 2026-04-18
Fix feed_team always resolving from home team perspective

lpukatch:
I realized the previous PR the home team was always considered the feed team. This fixes that.

I also added a default "National" fallback for non-feed channels. This way if you wanted to say "{feed_team) feed" it would resolve National if no team. Open to thoughts about this.

### DIFF
--- a/teamarr/consumers/event_epg.py
+++ b/teamarr/consumers/event_epg.py
@@ -405,16 +405,20 @@ class EventEPGGenerator:
             tvg_id = generate_event_tvg_id(event.id, event.provider, segment, exception_keyword)
             stream_name = stream.get("name", "")
 
-            # Build context using home team perspective
             # Inject exception_keyword into extra_vars so it resolves in all template fields
             keyword_value = exception_keyword if exception_keyword else ""
+            # When feed separation is active, build context from the feed team's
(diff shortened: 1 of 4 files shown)


egyptiangio:
Thanks @lpukatch — this PR pointed at a real problem, but I went after the deeper cause rather than patch the symptom.

Shipped on `dev` in `820cd67`: registry-level template-type scoping. Variables are now tagged TEAM_ONLY / EVENT_ONLY / ALL, and `GET /variables?template_type=…` filters the picker accordingly. Event templates no longer offer `{team}`, `{opponent}`, `{is_home}`, `{team_record}`, `{odds_moneyline}`, etc. — so the `team_id=event.home_team.id` hardcoding your PR fixed doesn't matter anymore; those variables simply aren't accessible on event templates.

`{feed_team}` keeps its empty-string graceful-disappear behaviour when no feed is assigned.

Closing. Appreciate the diagnosis.

egyptiangio:
One more note on the `feed_team → "National"` default: holding off on that. The empty-string behavior is intentional — not every channel without a detected feed is a national broadcast, so defaulting to "National" would write inaccurate metadata.

PR #211, opened 2026-05-18
Fix HockeyTech team schedule always returning empty results

sethwv:
I don't think he ever filed an issue, but this resolves an issue
Jonezy/Bootcamppanda was having where AHL/ECHL team-channels would not
generate EPG. This ended up being an issue with all hockeytech-provided-leagues.

The HockeyTech API returns `home_team` / `visiting_team` as integers,
but `team_id` is always a string (SQLite stores it as TEXT). Comparing
int to str is always False in Python, so no game ever matched and every
HockeyTech team channel produced no EPG.

Stringifies both sides before comparing, matching the pattern already
used in `get_team()` in the same file.

### DIFF
--- a/teamarr/providers/hockeytech/client.py
+++ b/teamarr/providers/hockeytech/client.py
@@ -318,7 +318,7 @@ class HockeyTechClient:
         team_games = []
         for game in schedule:
             # Check if team is home or away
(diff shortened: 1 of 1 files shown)


sethwv:
Still some testing to do on this one, but my dev/test instance is still occupied by #208 - will prove this out and mark ready for review in a few days.

PR #213, opened 2026-05-21
Serialize Dispatcharr token acquisition

bttfw:
## Summary

This PR makes Dispatcharr authentication caching safe under concurrent Teamarr API requests.

Root cause: `TokenManager` already shared token session state per Dispatcharr URL/user, but each `TokenManager` instance had its own lock. Concurrent short-lived clients could therefore authenticate in parallel. On top of that, `DispatcharrClient.close()` cleared the shared token, so connection/status checks could invalidate a token while other requests were actively fetching streams. When Dispatcharr returned HTTP 429, later callers retried authentication immediately and extended the throttle window.

## Changes

- Share the auth lock per Dispatcharr URL/user session.
- Preserve the shared token cache when closing a short-lived `DispatcharrClient`.
- Track Dispatcharr auth throttle responses via a shared `auth_retry_after` timestamp.
- Parse `Retry-After` or Dispatcharr's throttle detail text before retrying authentication.
- Add regression coverage for concurrent auth, throttle parsing, and close behavior.

## Validation

- `python -m pytest -q` -> 572 passed, 4 existing deprecation warnings.
- `python -m ruff check teamarr/dispatcharr/auth.py teamarr/dispatcharr/client.py tests/test_dispatcharr_auth.py` -> passed.
- Built a Docker image from this branch successfully.
- Read-only live pressure validation against a large Dispatcharr-backed library:
  - Test path: 108 Teamarr event groups fetched through the raw stream endpoint while polling Dispatcharr connectivity through Teamarr.
  - Before the final fix, the same read-only pattern produced repeated Dispatcharr auth throttling (`429`) and stream fetch failures.
  - After the fix: 108/108 group fetches succeeded, 0 operation errors, 0 auth throttle/error log lines during the run.
  - Stream-fetch operation timing after the fix: p50 88.3 ms, p95 203.0 ms, p99 258.8 ms, max 1129.0 ms, total 11.7 s.

p95/p99 here mean the 95th/99th percentile request durations, useful for showing tail latency rather than only the average case.

### DIFF
--- a/teamarr/dispatcharr/auth.py
+++ b/teamarr/dispatcharr/auth.py
@@ -5,7 +5,9 @@ and session management for Dispatcharr API integration.
 """
 
 import logging
+import re
 import threading
+import time
 from datetime import datetime, timedelta
 
 import httpx
@@ -32,6 +34,8 @@ class TokenManager:
     # Class-level session storage for multi-instance support
     # Key: "{url}_{username}" -> session dict
     _sessions: dict[str, dict] = {}
+    _session_locks: dict[str, threading.Lock] = {}
+    _registry_lock = threading.Lock()
 
     # Token refresh buffer (refresh this many minutes before expiry)
     TOKEN_REFRESH_BUFFER_MINUTES = 1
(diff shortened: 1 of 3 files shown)


egyptiangio:
Thanks for this — clean fix with solid test coverage and real validation numbers. Merged in v2.5.4.

One note for future contributions: we merge community PRs through `dev` first (not `main`) so they ride through our quality gates before hitting the release branch. I changed the base this time, but worth keeping in mind for next time. Appreciate the work!

PR #216, opened 2026-05-29
Stream Sorting Expansion

sethwv:
Add Home/Away Feed and team filter rules to stream ordering

## Summary
- Add `team_feed` and `not_team_feed` rule types that match streams based on broadcast perspective (home/away feed detection via team name + directional markers), with an inline league-grouped team picker; empty selection is a no-op
- Add optional team filter to the `stream_type: team` rule — selecting teams narrows matching to only those teams' streams (stored as `team|provider:league:id,...`); no selection matches all team streams
- Extract `VALID_RULE_TYPES` and `NO_VALUE_RULE_TYPES` as shared constants in `settings/types.py`, previously duplicated across the route, update, and read modules
- Extract `_find_matching_rule` helper eliminating duplicated catch_all loop logic from both `compute_priority` methods


egyptiangio:
Merged to `dev` 🎉 Thanks @sethwv!

On merge we added on top of your work:
- **Tests** — `tests/test_stream_ordering.py` (31 cases): team_feed/not_team_feed detection, the `stream_type` team filter, the `catch_all` fallback, 2-part/3-part key parsing, and no-connection degradation.
- **Hardening** — the team-term builder now drops generic stopwords (`the/and/for/with`) and sub-3-char city terms so feed patterns stay specific to the selected team.
- **Docs** — Stream Type, Home/Away Feed, and Everything Else rule types (plus team filters and feed-detection behavior) documented in the Channels settings guide.

Will ship in the next release.

PR #233, opened 2026-06-13
Add automatic international country name resolution for team matching

sethwv:
Streams from Spanish, Portuguese, French, and other-language providers use
localized country names ("Brasil", "Marruecos", "Suiza") that had no path
to the English team names ESPN uses. The only workaround was manually
creating entries in the team_aliases table.

This adds a CountryNameResolver that builds a locale-aware lookup table at
startup using pycountry and babel, covering ~250 countries across 20 locales.
It plugs in as a third tier in _resolve_alias(), after built-in TEAM_ALIASES
and user DB aliases, so user overrides are never affected.

FIFA-specific non-sovereign nations (Scotland, England, Wales, Northern
Ireland, Kosovo, Palestine, Chinese Taipei) and ESPN spelling quirks
(Turkey -> Turkiye) are handled via a hardcoded supplement since they fall
outside ISO 3166.

Tested against "Brasil vs Marruecos", "Suiza vs Catar", "Haiti vs Escocia",
and "Australia vs Turquia" -- all now resolve and match correctly without
any manual aliases.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,8 @@ dependencies = [
     "rapidfuzz>=3.0.0",
     "croniter>=2.0.0",
     "unidecode>=1.3.0",
+    "pycountry>=23.12.11",
+    "babel>=2.14.0",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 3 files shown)


egyptiangio:
Merged to `dev` — thanks @sethwv! Clean design: the locale map plugs in as a third alias tier after user overrides, and the normalization is consistent with the existing tiers (`canonical in tp.pattern`). Verified resolution for Brasil/Marruecos/Suiza/escocia/Turquía and added test coverage including the dependency-free fallback path.

PR #234, opened 2026-06-13
make sure home/away feed pattern is AFTER team to prevent false matches

lpukatch:
Prevent FEED TeamName from becoming a home/away channel. Changes to look for patterns after the teamname rather than just around


egyptiangio:
Merged to `dev` — thanks @lpukatch! The matchup-feed guard is a clean fix. Added two regression tests and tidied trailing whitespace on top of your commit.

PR #235, opened 2026-06-14
add mobile navigation

lpukatch:
Adds a mobile nav with hamburger menu and a fly in. 

Matches page will need to be looked at to keep everything with a mobile view-port but this PR adjusts the nav on mobile to prevent over scrolling

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -283,3 +283,19 @@ input[type="number"] {
   font-size: 0.6875rem;
   margin-left: 0.25rem;
 }
+
+/* Animations */
+.animate-slide-down {
+  animation: slide-down 0.2s ease-out forwards;
+}
+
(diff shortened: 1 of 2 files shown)


egyptiangio:
Merged to `dev` — thanks @lpukatch! Clean mobile nav that fixes the overscroll. Reconciled it with the concurrent nav rework and polished on top: the mobile menu now shows the first-run step badges (with visited drop-off) and Settings-as-step-0 to match desktop; raised the breakpoint md→lg since the full desktop nav still overflowed on ~768-1023px tablets; added aria-expanded/label on the toggle, close-on-Generate, and a tap-outside backdrop. The rest of the per-page mobile work you flagged (Matching, Settings/Channels grids, tables) continues under our mobile epic.

PR #236, opened 2026-06-14
 Support standard regex named groups in Custom Patterns

lpukatch:
This PR improves the Custom Regex experience by ensuring you can use standard regex syntax everywhere. The UI will now automatically convert your patterns into the backend-compatible format when saving, and convert them back when loading the form.
This ensures that what you test in the Pattern Tester is exactly what you see and edit in the form, without having to worry about syntax differences under the hood.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   teamarr:
-    image: ghcr.io/pharaoh-labs/teamarr:latest
+    build: .
     container_name: teamarr
     restart: unless-stopped
     ports:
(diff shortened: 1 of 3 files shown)


egyptiangio:
Merged to `dev` — thanks @lpukatch! Standard `(?<name>...)` syntax in the form is a real footgun-killer; verified the regex-utils round-trip. Dropped the docker-compose `build: .` line (local-dev only). Also folded in two related requests on top: the Pattern Tester 'Apply to Form' now also saves the group, the interactive team-regex builder is accent-inclusive (Atlético/München/Plzeň), and the pattern generator now emits JS `(?<name>...)` to match your new convention so generated patterns work in the JS tester.

PR #237, opened 2026-06-14
chore: optimize docker build and add gzip middleware

lpukatch:
use UV in the docker compose for deterministic builds
add on the fly gzip compression to the EPG output

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,10 +24,15 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     nano \
     && rm -rf /var/lib/apt/lists/*
 
-# Install Python dependencies from pyproject.toml (single source of truth)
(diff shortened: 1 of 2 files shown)


egyptiangio:
Merged to `dev` — thanks @lpukatch! Both halves landed, with two fixes:
- Generated + committed the missing `uv.lock` (pins current resolved versions); verified the image builds and runs on the venv (`docker build` + import check).
- gzip is in — confirmed `Content-Encoding: gzip` on XMLTV/JSON at runtime. Good news on the SSE worry: Starlette 1.2.0 already excludes `text/event-stream` by default, so the live progress streams aren't compressed — no exclusion code needed.
Dropped nothing else; CLAUDE.md now notes the `uv lock` workflow.

PR #238, opened 2026-06-14
Introduce a  fallback mechanism using stream's localized country names to determine the correct language channel 

lpukatch:
# Details
- Extends `CountryNameResolver` to natively capture the language corresponding to localized country names using the `pycountry` and `babel` locales.
- Extends the hardcoded `_FIFA_OVERRIDES` map to provide language fallbacks for Home Nations (e.g., Scotland, Wales) that are not considered sovereign ISO 3166 states.

# Examples
- A stream named `Peacock 06: Catar v. Suiza` has no language tag. The system matches "Suiza" through the `es` locale, resolving the language to "Spanish". It will then automatically flag and add the stream to the Spanish channel.
-  An explicitly tagged stream like `Qatar vs Suiza FHD (ENG)` continues to behave exactly as it did before. The explicit `(ENG)` exception keyword takes precedence and stops the fallback mechanism from triggering.

### DIFF
--- a/teamarr/consumers/lifecycle/service.py
+++ b/teamarr/consumers/lifecycle/service.py
@@ -636,6 +636,11 @@ class ChannelLifecycleService:
                             stream_name, conn
                         )
 
+                        # Fallback to language of country name if no explicit hint was found
+                        if matched_keyword is None and outcome.country_language:
+                            matched_keyword = outcome.country_language
+                            keyword_behavior = "separate"
+
                         # V1 Parity: If behavior is 'ignore', skip stream entirely
                         # This must happen BEFORE any channel lookup/creation
                         if keyword_behavior == "ignore":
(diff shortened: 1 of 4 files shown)


egyptiangio:
A few limitations to work through before this can merge:

**1. Needs to be opt-in.** It fires for every matched stream globally, but feed separation is otherwise gated behind a setting — this would start creating language-suffixed channels for everyone on update. Suggest a toggle.

**2. It separates English streams.** Running the resolver against the branch: `England`/`Scotland`/`Wales` resolve to `"English"`, and — worse — common English country names collide with other locales, so `Argentina` → `Norwegian` and `Qatar`/`Australia` → `Finnish`. So a plain-English `Argentina vs Australia` gets split into a "Norwegian"/"Finnish" channel. The default/primary language probably shouldn't trigger separation at all.

**3. Unreliable when names are spelled the same/similarly across languages.** `_lang_map` is overwritten per-locale (last-writer-wins in `_LOCALES` order), so identical spellings (`Portugal`, `Brasil`, `Argentina`) resolve to whichever locale comes last rather than the real broadcast language. Better to only assign a language when a spelling maps *unambiguously* to one locale, or use a deliberate priority order.

**4. The core premise is shaky.** Even with perfect resolution, a localized country name in the title isn't a reliable indicator of the actual broadcast language. Provider naming is inconsistent — a title can use localized team names while the feed is in another language (and vice versa) — so this risks mislabeling streams even when the name→language mapping is correct.

lpukatch:
totally right on this. I think this should maybe be an event group toggle rather than a global one. I'll rescope and adjust based on your feedback

Issue #29, opened 2025-12-09
[Feature]: Motorsports Compatibility


bradarussell:
### Problem / Use case

Add the ability to follow different auto racings series in a similar manner to following a team.  Examples are:

- NASCAR Cup Series
- Formula 1
- NASCAR O'Reilly Auto Parts Series (formerly the Xfinity Series)
- NASCAR Craftsman Truck Series
- MotoGP

### Proposed solution

With the ability to add a series similarly to a team, there are also a variety of event types that occur on a race weekend, such as: 

- Preseason Testing
- Practice sessions
- Qualifying
- Pre-race
- Race
- Post-race

Teamarr would lookup and identify listings determined by:

- Series
- Event
- Session Type
- Live or Recorded

### Additional context

The events are typically listed by a full name, with formats similar to:

- NASCAR: Autotrader 400, Echopark Speedway, 2/22/2026
- F1: Luis Vuitton Astralian GP

ESPN has schedule links but they don't have 2026 data yet:
- NASCAR Cup Series: https://www.espn.com/racing/schedule
- Formula 1: https://www.espn.com/f1/schedule
- NASCAR O'Reilly Auto Parts Series: https://www.espn.com/racing/schedule/_/series/xfinity (pending change with series name change)
- NASCAR Craftsman Truck Series: https://www.espn.com/racing/schedule/_/series/truck

The 2026 schedules are available on the series sites:
- NASCAR Cup Series: https://www.nascar.com/nascar-cup-series/2026/schedule
- Formula 1: https://www.formula1.com/en/racing/2026
- NASCAR O'Reilly Auto Parts Series: https://www.nascar.com/nascar-oreilly-auto-parts-series/2026/schedule
- NASCAR Craftsman Truck Series: https://www.nascar.com/nascar-craftsman-truck-series/2026/schedule

tomwinterrose:
I have this working, I can create a PR if you'd like. Just doing a little more testing/polishing

egyptiangio:
Please hold off until the major UI/UX redesign currently underway, and rebase against that for review. Will make everyone's life a lot easier :) . Hopefully in the next day or two.

tomwinterrose:
You got it!

egyptiangio:
[...]

PR #242, opened 2026-06-15
Add motorsports support (F1, NASCAR, IndyCar, WEC, IMSA)

tomwinterrose:
## Summary
- Adds racing/motorsport support as an additive feature on top of the existing team-vs-team and event-card matching pipeline: new `RacingSession`/`RacingResult` data model, ESPN racing provider (F1 fully verified reference league, plus NASCAR Cup/Xfinity/Truck and IndyCar), per-session EPG segment expansion, racing stream classification/matching, and a new "motorsports" template variable category with F1 sample data.
- Switches WEC and IMSA to TheSportsDB (TSDB) with full session grouping (FP1-3, class-specific Qualifying/Hyperpole, Race, Prologue AM/PM for WEC; single-event rounds for IMSA), including per-event/per-league session duration handling (e.g. 24 Hours of Le Mans, Twelve Hours of Sebring).
- A series of pipeline fixes found via end-to-end testing: cache serialization preserving session/circuit data, multi-session event status/lifecycle anchoring on the last session rather than the first, and fixed session segment durations.
- Disables MotoGP (`leagues.enabled = 0`) pending a future TSDB-backed migration (idLeague 4407) — ESPN's `racing/motogp` scoreboard endpoint returns no usable data. League count docs updated 133 -> 132.
- Adds TSDB badge logos for IMSA/WEC and fixes broken ESPN logo URLs for NASCAR series and IndyCar.

## Test plan
- [x] `ruff check teamarr/` — clean
- [x] `pytest tests/ -v` — passes (same pre-existing unrelated failures in `test_country_resolver.py`/`test_stream_ordering.py` as on `dev`)
- [x] `cd frontend && npm run build` — succeeds
- [x] Manual: verified F1 event matching/EPG segment generation, WEC/IMSA TSDB-sourced multi-session events, and MotoGP correctly hidden from league picker (134/135 leagues loaded)


egyptiangio:
Thanks for this @tomwinterrose — it's a huge contribution (F1/NASCAR/IndyCar via ESPN, IMSA/WEC via TSDB, the whole racing model + per-session EPG expansion). The code is clean and well-tested, so the plan is to **merge it for credit and do a few small follow-up fixes on `dev`** rather than ping-pong the branch.

One thing I wanted to check with you before we touch it — the **TSDB tier split**: `imsa` is marked `tsdb_tier='free'` while `wec` is `'premium'`. From the racing provider, both leagues' schedules are fetched only via `eventsseason.php` (the free `eventsnextleague.php`/`eventsday.php` return "Invalid League ID" for these IDs), and `eventsseason.php` is a premium endpoint — so I'd have expected **both** to be `premium`. As marked, a user without a premium TSDB key wouldn't have IMSA skipped in prewarm and would get silently-empty IMSA guides.

Was there a specific reason you set IMSA to free — e.g. did you confirm idLeague 4488 resolves on a free key where WEC (4413) doesn't? If it was intentional we'll keep it; if it was an oversight we'll flip it to `premium` in the follow-up. Just don't want to 'fix' something you found to be deliberate.

Thanks again — this closes #29 and lands motorsports support. 🏁

tomwinterrose:
The only reason was that IMSA had few enough sessions in TSDB that a premium key wasn't required. But I think it probably is cleaner to have both require premium. IMSA isn't nearly as popular as the the series covered by ESPN, so I doubt many users would be impacted by making it gated behind premium.

PR #259, opened 2026-06-21
Motorsports support

tomwinterrose:
Added a fix for NASCAR matching and updated NASCAR Xfinity to the current title sponsor


egyptiangio:
Superseded by #263, which landed on dev as a take-and-fix (`3749926`) — see the writeup there. Thanks for both!

PR #261, opened 2026-06-22
Cut redundant ESPN scoreboard fetches during generation

lpukatch:
## Summary

Generation was hitting ESPN's scoreboard endpoint far more than necessary — the same `(league, date)` scoreboard was being fetched once per team, once per racing thread, and once as a redundant pre-flight before matching even ran. This branch collapses all of that onto a single shared, deduplicated per-`(league, date)` events cache.

Three independent fixes, no behavior change to matching coverage or EPG output:

## 1. Share scoreboard fetches across team-schedule scans

Team-schedule generation scanned each future day's full-league scoreboard once **per team** (`days_ahead × N_teams` fetches), bypassing the cached, league-wide `get_events`. `SportsDataService` now injects `get_events` into the ESPN provider so the future-day scan reuses the shared per-`(league, date)` cache — each day's scoreboard is fetched once per run and shared across every team and event-group match.

- **Measured: 84 → 14 fetches (83% fewer)** for 6 teams × 14 days; flat at `days_ahead` regardless of team count.
- Falls back to direct per-day fetches when used standalone (no cache wired).

## 2. Single-flight dedup for concurrent identical fetches 

Parallel stream matching fans out across threads that frequently want the same `(league, date)` scoreboard. Without coordination, every thread racing past the cache miss issued its own upstream fetch, and the coalesce marker had a check-then-act race.

## 3. Drop the redundant pre-flight event fetch

`_fetch_events` did a full multi-day scoreboard sweep before matching, but its result was only used for an "are there any events?" early-exit guard — it was **never matched against**. The matcher fetches its own events (30-day window, past days cache-only), so the sweep was pure overhead, and its 7-day lookback existed solely to stop the guard from false-bailing mid-week for weekly sports (NFL).

image

### DIFF
--- a/teamarr/consumers/event_group_processor/__init__.py
+++ b/teamarr/consumers/event_group_processor/__init__.py
@@ -3,18 +3,17 @@
 Connects stream matching to channel lifecycle:
 1. Load group config from database
 2. Fetch M3U streams from Dispatcharr
-3. Fetch events from data providers (parallel with ThreadPoolExecutor)
-4. Match streams to events
-5. Create/update channels via ChannelLifecycleService
-6. Generate XMLTV EPG
-7. Optionally push EPG to Dispatcharr
+3. Match streams to events (the matcher fetches events itself, per league)
+4. Create/update channels via ChannelLifecycleService
+5. Generate XMLTV EPG
+6. Optionally push EPG to Dispatcharr
 
 This is the main entry point for event-based EPG generation.
 
 Package layout (iua3.7):
 - processor.py       — EventGroupProcessor coordinator + convenience functions
 - results.py         — result dataclasses (processing/batch/preview/enforcement)
-- stream_fetcher.py  — M3U stream fetching/filtering + provider event fetching
+- stream_fetcher.py  — M3U stream fetching/filtering + known-league lookup
 - matching.py        — stream→event matching, EPG index, feed/segment expansion
 - team_filter.py     — team include/exclude filtering + filtered-channel cleanup
 - persistence.py     — matched/failed stream persistence for run analysis
@@ -37,10 +36,8 @@ from .results import (
     PreviewStream,
     ProcessingResult,
 )
-from .stream_fetcher import MAX_WORKERS
 
 __all__ = [
-    "MAX_WORKERS",
     "SEASON_POSTSEASON",
     "BatchProcessingResult",
     "EnforcementStepResult",
--- a/teamarr/consumers/event_group_processor/processor.py
+++ b/teamarr/consumers/event_group_processor/processor.py
@@ -656,33 +656,11 @@ class EventGroupProcessor(
                 )
                 return result
 
-            # Step 2: Fetch events from data providers
-            # Use subscription leagues (per-group override → global fallback)
+            # Step 2: Resolve subscription leagues (per-group override → global
+            # fallback). The matcher fetches its own events with a 30-day window,
+            # so there is no separate pre-flight event fetch here — an eventless
+            # group simply produces zero matches downstream.
             effective_leagues = self._get_subscription_leagues(conn, group)
-            events = self._fetch_events(effective_leagues, target_date)
-            logger.info(
-                f"Fetched {len(events)} events for group '{group.name}' leagues={effective_leagues}"
-            )
-
-            if not events:
-                result.errors.append(f"No events found for leagues: {effective_leagues}")
-                result.completed_at = datetime.now()
-                stats_run.complete(status="completed", error="No events found")
-                save_run(conn, stats_run)
-                # Update stats - streams are eligible but no events to match against
-                update_group_stats(
-                    conn,
-                    group.id,
-                    stream_count=result.streams_after_filter,  # Eligible streams
-                    matched_count=0,
-                    filtered_stale=filter_result.filtered_stale,
-                    filtered_include_regex=filter_result.filtered_include,
-                    filtered_exclude_regex=filter_result.filtered_exclude,
-                    failed_count=result.streams_after_filter,  # All unmatched due to no events
-                    filtered_not_event=filter_result.filtered_not_event,
-                    total_stream_count=result.streams_fetched,
-                )
-                return result
 
             # Step 3: Match streams to events (uses fingerprint cache)
             match_result = self._match_streams(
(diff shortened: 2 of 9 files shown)


lpukatch:
This needs some testing to make sure there's no unintended side effects but wanted to get this out so it's visible. I'll do some testing in the upcoming weeks

egyptiangio (commented):
Thanks for this, @lpukatch — really strong PR. Correct, well-commented, cleanly separated into three fixes, and it genuinely cuts the fetch volume. I merged it onto current `dev` locally and ran the full suite, and went through the single-flight code by hand — the concurrency is *right*. Recommending merge after one required rebase and a couple of confirmations below.

## 🔴 Blocking — rebase needed (conflicts with just-merged #287)

`dev` moved under you: #287 landed a per-run fetch-count regression guard that replaced the `_DictCache` class in `tests/providers/test_event_negative_cache.py` with a shared `FakeCache` in `tests/fakes.py`. Your PR adds a `lock_key` context manager to `_DictCache`.

Heads-up: git auto-merges these **without a conflict marker**, but the result is broken Python — an orphaned `@contextmanager def lock_key` with no enclosing class → `IndentationError`, and the whole suite fails to collect. GitHub's green "mergeable" is a false positive here.

Fix on rebase (I verified this takes the full suite green — 1383 passing):
- Add `lock_key` to the shared `FakeCache` in `tests/fakes.py` instead:
  ```python
  from contextlib import contextmanager
  @contextmanager
  def lock_key(self, key):
      yield
  ```
- Drop the edit to `test_event_negative_cache.py` (it uses `FakeCache` now).

This matters beyond that one file: production `sports_data.py` now calls `self._cache.lock_key(...)` in **both** `get_events` and `get_event`, so every test cache stand-in needs `lock_key` — including the one behind #287's new `TestScoreboardFanout` tests, which `AttributeError` without it.

## 🟡 Should address

**1. Confirm eventless-group stats parity.** Removing `_fetch_events` also removed the early-exit that, on zero events, wrote a specific `update_group_stats(matched_count=0, failed_count=<all eligible streams>, …)` and `stats_run.complete(error="No events found")`. The new flow falls through to `_match_streams`, which matches nothing — functionally fine, but could you confirm the downstream zero-match path writes the **same** group stats (failed/filtered counts, total, run status)? Just want to be sure the dashboard/history doesn't silently change for eventless groups — it's the one behavioral change I couldn't fully confirm by reading.

**2. New Pyright errors from the `_CACHE_MISS` sentinel.** Typing `load_from_cache() -> list[Event] | object` leaks `object` into the callers — Pyright flags `home_team`/`away_team` access and the `_enrich_event_teams(object)` arg (sports_data.py ~362–363, 396, 408). Pyright is advisory here and the repo has pre-existing errors (the `_client` ones aren't yours), but this adds to the pile. A sentinel class narrows cleanly with `isinstance`:
```python
class _CacheMiss: ...
_CACHE_MISS = _CacheMiss()
def load_from_cache() -> list[Event] | _CacheMiss: ...
```

## 🟢 Nits (non-blocking)

- **Document the non-reentrant invariant on `lock_key`.** It's a plain `threading.Lock`, which is correct — the refcounting is sound and incrementing before `acquire()` prevents the delete-under-waiter race. But a fetch path that re-entered `get_events`/`get_event` for the *same* key would self-deadlock. A one-line "callers under the lock must not re-enter for the same key" would future-proof it. (Today's injection is safe — the cached fetcher is only used in `get_team_schedule`, which isn't lock-wrapped and hits a different key.)
- **`lock_key` `finally` re-fetches `entry` from the map** rather than using the captured local. Harmless (your own refcount keeps it alive, so `is not None` never fires), but the captured `entry` reads clearer.
- **Stale docstring:** `stream_fetcher.py`'s class docstring still says "and events from data providers" after the fetch method was removed.
- **Worker cap on the matching path:** the removed `_fetch_events` copy of `ESPN_MAX_WORKERS` is gone (the env var itself is fine — still honored in `team_processor.py`/`cache/refresh.py` + docs). Single-flight makes an uncapped matcher much less risky, but worth confirming the matcher's fan-out still has *some* cap for DNS-throttled users (PiHole/AdGuard).

## ✅ Checked and NOT problems (so you don't have to worry about these)
- `event_match_days_ahead` is still honored (`matcher.py:290`), and `event_match_days_back` supersedes the old hardcoded 7-day NFL lookback — no window regression.
- `ESPN_MAX_WORKERS` isn't orphaned (used in two other consumers + docs).
- Double-checked locking in both methods is correct; `_CACHE_MISS` properly distinguishes a cached empty `[]`/negative-`None` from a true miss.

Nice work overall — the measured 83% reduction lines up with what the #287 call-count tests show. Ping me once rebased and I'll re-run.

lpukatch:
@egyptiangio this should be all good to review now

egyptiangio (approved):
Re-reviewed after the rebase — **approving**. All items from the previous review are addressed:

- ✅ Rebase conflict resolved exactly as suggested (`lock_key` on the shared `FakeCache`; no per-file test edit) — full suite green in CI including #287's fanout tests
- ✅ `_CacheMiss` sentinel class (with `__slots__`, nice touch) — Pyright advisory stays at 0 on the merge ref
- ✅ Non-reentrancy documented on `lock_key`, `finally` now uses the captured entry, stale docstring fixed

And the two open confirmations, now verified on my side:

1. **Eventless-group stats parity — confirmed equivalent.** The fall-through path's post-matching `update_group_stats` writes the same values the old early-exit wrote (matched=0, failed=all eligible, same filter/total counts) plus fields that are zero for an eventless run. Only delta: the run no longer records `error="No events found"` — dashboard/run-history shows a clean completed run instead of an error annotation for eventless groups. I'd call that more correct, not a regression.
2. **No uncapped fan-out replaced the removed pool.** The matcher's event prefetch is a sequential per-league loop with its own cache-only rules and a cross-group shared-events layer, so DNS-throttled setups (PiHole/AdGuard) aren't exposed; `ESPN_MAX_WORKERS` remains honored where pools still exist.

One non-blocking observation for a future pass: `SportsDataService.__init__` injects `set_cached_events_fn` into constructor-passed providers, but `add_provider()` doesn't — no current callers use `add_provider`, so it's latent, but worth mirroring the injection there someday to avoid a surprise.

Really strong work — measured 83% fetch reduction, correct concurrency, and the review turnaround was fast and thorough. Merging.

PR #263, opened 2026-06-25
motorsports matching issues

tomwinterrose:
Add racing_league_hints fallback, venue.country to matching logic

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,7 +50,7 @@ EPG:     Kansas City Chiefs @ Philadelphia Eagles
 ## Features
 
 - **350+ leagues across 14 sports** - Football, basketball, hockey, baseball, soccer (~250 leagues via ESPN discovery), cricket, lacrosse, MMA, boxing, rugby, volleyball, Australian football, softball, and racing (F1, NASCAR, IndyCar). 134 pre-configured leagues plus dynamically discovered soccer leagues.
-- **226 template variables** - Customize channel names and EPG with team records, scores, venues, broadcasts, standings, playoff status, motorsports sessions/results, and more
+- **232 template variables** - Customize channel names and EPG with team records, scores, venues, broadcasts, standings, playoff status, motorsports sessions/results, and more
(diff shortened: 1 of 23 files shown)


egyptiangio:
Thank you for this — the NASCAR provider is exactly the right call (ESPN only exposes the single race competition, and the official API's full weekend sessions make the racing matcher's date-coverage work properly). This has landed on `dev` as a take-and-fix (`3749926`), with your work credited in the commit.

**What we took as-is:** the NASCAR provider core + session parsing, the classifier/is_racing hint triggers, the EPG racing fallback for mixed groups, the AT/city guard, the 6 race-format variables, the O'Reilly rename, and your test suites (nice work on those — parametrized session mapping and the stubbed-fetch builder made review easy).

**What we fixed on top:**
- `_match_single` had a leftover duplicate `_route_to_outcomes` call — every stream ran the full match pipeline twice.
- The three new `Event` fields weren't serialized in `provider_cache.py`, so the lap/stage variables rendered empty for cache-served events (the normal generation path).
- The provider fetched the season in its constructor — a startup-time API blip left it permanently empty until restart. Rebuilt as lazy-load with a 6h TTL, 15m empty-retry, and keep-old-data-on-failed-refresh.
- `venue.country` as a peer fuzzy candidate could outrank a real name match on another event (token_set_ratio scores a bare country subset at 100) — demoted it to a single-event sanity backup + unique-country fallback, which preserves your "at Mexico City" use case.
- `_RACING_LEAGUE_HINTS` now has a test that derives the expected set from `schema.sql`, so future racing leagues can't silently miss it.

**What we carved out:** `StaticCalendarProvider` — no league maps to `provider='static'` so it was unreachable, the seed calendar's own comment says the dates are guesses, and we removed a static provider once before (hand-maintained season calendars are a maintenance model we've decided against). IMSA/WEC stay on TSDB. Also dropped the IMSA free-tier doc change — TSDB leagues are premium-tier by policy regardless of event volume.

Ships in the next release. Closing #259 as superseded by this one. Thanks again!

Issue #271, opened 2026-07-04
[League Request]: FIBA Basketball


sultan7swing:
### Sport

Basketball

### League/Competition Name

FIBA Basketball

### Common Abbreviation

FIBA

### Country/Region

World Wide,

### Competition Level

Professional (top tier)

### Data Sources

https://www.espn.com/fiba/
https://www.fiba.basketball/en

### Teams/Participants

All countries 

### Season Information

All year around

### Additional Context

It's an international basketball league, the one ruling international basketball

egyptiangio:
Landed on dev (`0882695`, via PR #272 by @lpukatch): FIBA Basketball World Cup (men's and women's) via TheSportsDB — note these are **premium-tier TSDB leagues** (require a TSDB premium key in Settings) because qualifier volume exceeds the free tier's daily cap — plus the Australian NBL via ESPN as a bonus. Ships in the next release; closing then.

egyptiangio:
Shipped in [v2.9.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.9.0) — closing.

PR #272, opened 2026-07-04
feat(leagues): add FIBA World Cup, Australian NBL, Olympic basketball

lpukatch:
- Add Australian NBL via ESPN (basketball/nbl) — verified against live 2026-27 schedule data (real fixtures, e.g. Cairns Taipans @ Perth Wildcats)
- Add FIBA Basketball World Cup (M/W) via TSDB, premium tier — ESPN's basketball/fiba endpoint only tracks the tournament final and returns nothing between World Cups; TSDB tracks year-round qualifiers instead (confirmed live games today). Gated premium since parallel confederation qualifiers can exceed the free tier's 5-events/day/league cap.

Closes: #271

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -49,7 +49,7 @@ EPG:     Kansas City Chiefs @ Philadelphia Eagles
 
 ## Features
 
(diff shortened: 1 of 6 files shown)


egyptiangio:
Taken onto dev with credit as `0882695` — thank you! Everything checked out on review: ESPN `basketball/nbl` returns real 2026-27 fixtures and 10 importable teams, TSDB ids 4549/4891 resolve to exactly the right leagues, and the premium gating matches our TSDB tier policy (nice catch on the parallel-qualifier volume).

The branch had drifted into conflict with dev (the tennis + NASCAR leagues landed in the same schema region), and the league counts moved underneath it — dev is at 140 pre-configured / 93 ESPN now — so I applied it directly with recounted docs rather than round-tripping a rebase. Closing this PR; #271 closes at the next release, which this will ship in.

Issue #264, opened 2026-06-28
[Bug]: Only one stream added to channel with multiple identically named streams from same provider with different logins


FractalBoy:
### Tag / Commit

2.8.0

### What happened?

I have multiple logins for the same provider with separate M3Us for each. It seems that teamarr is only picking one stream per provider instead of one stream per M3U. 

### Expected behavior

Add all matching streams to the channel. 

### Steps to reproduce

1. Add two or more M3Us in Dispatcharr from the same provider (you could probably just add an exact duplicate to reproduce)
2. Add groups for both to Teamarr
3. Generate channels - only one M3U's streams are used. 

### Logs

```shell
I'll work on getting relevant logs.
```

FractalBoy:
Logs from the last EPG generation

[teamarr.log](https://github.com/user-attachments/files/29442441/teamarr.log)

egyptiangio:
Fixed on dev by PR #269 (`55f8e15`) — matched-stream lookups were keyed by stream name, so identically named streams from multiple logins collapsed to one. Lookup is now ID-first; each account's stream attaches to the channel. Will ship in the next release.

egyptiangio:
Shipped in [v2.9.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.9.0) — closing.

FractalBoy:
@egyptiangio this is working now but there is a visual bug. Under managed channels, if you look at the streams under a channel, it will list the same stream with the same name and account multiple times. Seems to be correct in Dispatcharr though.

egyptiangio:
@FractalBoy thanks for the follow-up — confirmed and diagnosed. The attach fix stores each stream's account **id** correctly, but the display **name** was never resolved per-stream: every stream in a group gets labeled with the group's configured account name, so your two logins' streams render as identical rows (Dispatcharr is right; Teamarr's label is wrong). It can also affect m3u-type stream-ordering rules. Tracking the fix in #297 — will land on dev shortly and ship in the next release.

egyptiangio:
@FractalBoy the fix is on dev (#297, PR #298) and verified live — identically named streams from different logins now each show their own account in Managed Channels, and rows written before the fix self-correct on the next EPG generation run. It'll ship in the next release.

PR #298, opened 2026-07-09
fix(lifecycle): resolve M3U account name per stream, not per group (#297)

### LINKED ISSUE #297: [Bug]: Managed Channels shows wrong/duplicate M3U account name for multi-login streams (follow-up to #264)
## Report
From @FractalBoy on #264 (after the v2.9.0 fix): the multi-login attach fix works, but the Managed Channels stream list shows the same stream name **and the same account** repeated, while Dispatcharr shows the streams correctly distinguished by account.

## Root cause
The per-stream account **name** is never resolved — only the id is:

- `event_group_processor/stream_fetcher.py` builds stream dicts with `m3u_account_id` but no `m3u_account_name`
- `lifecycle/creator.py` (consolidate path, ~L541) does `stream.get("m3u_account_name") or group_config.get("m3u_account_name")` — the first arm is always `None`, so **every stream in a group is labeled with the group's single configured account name**
- the create-new-channel path (~L901) uses `group_config.get("m3u_account_name")` directly
- the UI (`ManagedChannelsTable.tsx`) displays `m3u_account_name` per stream row → identical rows for what are actually distinct streams from different logins

Not purely cosmetic: the same wrong name feeds `compute_stream_priority_from_rules`, so **m3u-type stream-ordering rules mis-evaluate** for multi-login setups.

## Fix
- Build an account id→name map in `stream_fetcher` via `m3u_manager.list_accounts()` and stamp `m3u_account_name` on each stream dict
- Prefer the stream dict's account name in both creator paths
- Self-heal existing rows: refresh stored `m3u_account_name` on generation when it differs from the stream's actual account

Refs #264.

egyptiangio:
## Summary
Fixes #297 (follow-up to #264, reported by @FractalBoy): Managed Channels showed identically named multi-login streams with the same account label, because the per-stream account **name** was never resolved — only the id. Every stream in a group was labeled with the group's single configured account name, which also fed wrong input to m3u-type stream-ordering rules.

- `stream_fetcher.py`: resolve account id→name via `m3u_manager.list_accounts()` (cached per processor instance, tolerant of API failure) and stamp `m3u_account_name` on every stream dict (both the M3U-group path and the channel-source path)
- `lifecycle/creator.py`: prefer the per-stream account name in both the consolidate-attach and create-new-channel paths (group config stays as fallback)
- Self-heal: new `update_stream_account_name` DB helper refreshes the stored name/id on already-attached rows each generation run, so rows written before this fix correct themselves without user action (guarded no-op when values already match; never nulls on a transient account-list failure)

## Test plan
- [x] 10 new tests in `tests/lifecycle/test_stream_account_name.py`: DB helper heal/no-op/COALESCE/removed-row guards; fetcher stamping for multi-login duplicates; per-processor caching (custom accounts included); API-failure tolerance
- [x] `pytest tests/ -v` — 1389 passed
- [x] `ruff check teamarr/ tests/` — clean
- [x] `cd frontend && npm run build` — clean (no frontend changes)


egyptiangio:
**Live verification on the dev instance** (which exhibited the exact reported symptom — two identically named streams both labeled "IPTV - FlareTV Onyx"): triggered a full EPG generation with this branch running, and every multi-login stream pair now shows its own account, e.g.

```
Atlanta Braves at Pittsburgh Pirates | 12:35 PM ET | account: IPTV - FlareTV Onyx
Atlanta Braves at Pittsburgh Pirates | 12:35 PM ET | account: IPTV - Aura
```

Generation completed with no errors; 14 channels with multi-account streams all correctly attributed.

Issue #329, opened 2026-07-10
Best-in-class default templates, Gracenote-modeled (epic teamarrv2-tvnk)


egyptiangio:
Umbrella issue for the roadmap epic `teamarrv2-tvnk` (predates the issue-first workflow; filing now that implementation begins).

Produce best-in-class default templates modeled on authoritative Gracenote EPG data: real Gracenote samples as the target, probed real-world EPG as the approximation, the delta defining the work. Ship a curated seeded set (per sethwv's practice + our gracenote_category work), with recommended scoping, relative art paths (z02s), and new variables where Gracenote fidelity demands them (article-aware team naming, national-team detection — see bead `tvnk.7`).

Now includes tennis (bead `tvnk.13`) since epic mf7 made it a supported sport.

Work proceeds child-bead by child-bead under the epic: `bd show teamarrv2-tvnk`.

egyptiangio:
First slices on dev (b59ac44): curated 6-template Starter set (#330, #331 — replace-pristine seeding, live-verified reference-safe on a production DB; tennis included; abbrev fallback retired No-Abbrev; MiLB kept pending Gracenote branding decision), plus the Gracenote tmsapi capture reference (#332 — universal pattern, tennis validated, MiLB='Minor League Baseball', ~48h API lookback only). Epic continues: tvnk.14 (ESPN-copy-first descriptions incl. postgame) is next.

egyptiangio:
tvnk.14 (ESPN-copy-first descriptions) landed on dev: `a4ffccd` (PR #341). New `has_recap`/`has_preview` conditions, filler description fallthrough (defines the postgame_fallback vs postgame_conditional precedence), starter set rewired copy-first — incl. the Tennis empty-recap fix and Default Team's .last-suffixed postgame. Remaining in the epic: tvnk.7 (article-aware naming vars), tvnk.15 (Tier-2 structured previews), tvnk.8/.9/.12/.4.

egyptiangio:
Follow-up landed on dev: `365488f` (PR #342) — the Conditions tab is now visible for event templates, so the seeded has_preview rows (and any user-added event conditionals) are editable in the UI.

egyptiangio:
[...]

PR #331, opened 2026-07-10
feat(templates): starter-set v2 — (Starter) names + abbrev fallback (#329)

egyptiangio:
## Summary
Maintainer feedback on #330's set:

- **"(Starter)" parentheticals** on all 6 members — the included-with-Teamarr origin is visible in every list/picker/export. Unedited rows from the first iteration rename **in place** (same id; live-verified: a production DB's 7 references all survived two healing generations).
- **No-Abbrev Event retired**: the four `*_team_abbrev` variables now fall back `abbreviation → short_name → name` (only real abbrevs uppercased → `EPL | Man United/Arsenal`), so Default Event covers abbrev-less leagues. The retired member is removed only when it's still our unedited seed.
- **MiLB kept** deliberately — folding it is an alias/branding decision better made after the Gracenote analysis (tvnk.9/12).
- **Hint link fixed**: the "recommended scoping guide" link 404'd (docs site publishes from main at release time). The hint is now self-contained with an inline scoping summary.

## Test plan
- [x] 14 seeding tests (4 new: prior-name rename-in-place, retired-member removal, edited-retired safety, abbrev fallback rendering)
- [x] 1420 passed · ruff · Pyright clean · frontend build clean
- [x] Live-verified on production DB: 6 templates, ids stable across rename, references intact; hint renders inline (correctly counting only unassigned starters)

### DIFF
--- a/docs/guide/templates/defaults.md
+++ b/docs/guide/templates/defaults.md
@@ -24,13 +24,16 @@ same assignments, fixed art.
 
 | Template | Type | Designed for | Channel name style |
 |----------|------|--------------|--------------------|
-| **Default Team** | team | Any team channel | `{team_name}` |
-| **Default Event** | event | US pro leagues with abbreviations (NBA, NFL, MLB, NHL…) | `NBA \| DET/LAL` |
-| **Combat Event** | event | UFC / PFL / boxing (card segments) | `UFC 310 Main Card` |
-| **International Event** | event | National teams, international competitions | `NED v JPN` |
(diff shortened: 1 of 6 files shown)


egyptiangio:
Added two live-found UX fixes: (1) assignment mutations now invalidate the templates list, so the Usage column updates in real time; (2) Usage chips resolve league/sport slugs to display names (`fifa.world` → "World Cup", `mma` → "MMA"). Both verified live.

Issue #262, opened 2026-06-24
[Feature]: Boolean logic for Stream priority rules


smevawala:
### Problem / Use case

I would love the ability to have some boolean logic in the stream priority rules.  It already exists for stream stats, but I would like the ability for prioritizing stream name matching with preferred M3U accounts.

(I basically want all epg based streams to be last, even if they come from my preferred m3u account)

### Proposed solution

Add a "AND" condition to all the stream priority rules, exactly how its done with stream stats,

"Or" already could be done by adding 2 rues with the same priority.
 A "Not" may be useful too, have not really thought that thru

lpukatch:
Tracked as bead `teamarrv2-r8e2` — Boolean logic (AND/NOT) for stream priority rules, mirroring the existing boolean support in stream stats rules.

egyptiangio:
Landed on dev via #336 (merge `2367855`) — additive scoring for Stream Priority by @lpukatch. Closes at the next release.

egyptiangio:
Shipped in [v2.10.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.10.0).

PR #336, opened 2026-07-10
feat(stream-ordering): additive scoring for Stream Priority (#262)

lpukatch:
## Summary

Reworks Stream Priority ordering from first-match-wins into a two-class model (Radarr Custom-Formats style), implementing epic `teamarr-5ag` and addressing #262.

- **Scoring rules** (`mode="score"`, default for new rules) — additive signed **points**; a stream sums every scoring rule it matches, highest total sorts first. Negative points demote.
- **Priority rules** (`mode="priority"`, hard bands) — the legacy first-match-wins behaviour, kept as a strict "must always win/lose" escape hatch. Bands always outrank scoring.
- Sort key per stream: **band → total score within band → added-at**. Pure-priority rulesets skip the score scaling entirely, so existing configs are **byte-identical**.

## How this addresses #262

The issue asks for boolean AND/NOT on priority rules; the motivating case is *"keep EPG-matched streams last, even from a preferred M3U account."* The additive model handles that directly — e.g. `M3U +100` and `EPG matched −100000` nets negative so the EPG stream sinks below everything, while non-EPG streams from that account stay boosted. Signed points + hard bands cover the AND/NOT use cases without a per-rule boolean operator.

## Changes

- Data model: `mode` + signed `points` on ordering rules; legacy rows default-coerce to `priority`/`0`.
- `StreamOrderingService`: band + additive-score sort key; `evaluate_rules` explainer reports band winner + score contributors.
- Settings API: `mode`/`points` exposed and validated.
- UI: two sections (**Scoring** primary, **Priority** hard-order) with signed-points input, optional "Everything Else" baseline, and a clickable priority-explainer popover on the Channels page.
- Docs: rewrote `docs/guide/channels/stream-priority.md` for the two-class model.

## Migration

None required. Rules live in a single JSON settings blob (no schema column); legacy blobs deserialize as priority-mode rules and produce identical ordering. Old exports import cleanly.

## Testing

- `pytest tests/` — **1445 passed**
- `cd frontend && npm run build` (tsc + vite) — clean

Closes epic `teamarr-5ag` (children `5ag.1/.3/.4/.5/.6/.7`) and reconciles bead `teamarrv2-r8e2`.


egyptiangio:
Triage: reviewed the design summary — the two-class model (additive scoring + legacy priority bands as an escape hatch) is a clean answer to #262's AND/NOT ask, and byte-identical sorting for pure-priority configs is exactly the right compatibility bar. All checks are green. Full review queued behind the v2.9.2 release going out today; expect detailed review within ~24h. Bead teamarrv2-r8e2 covers #262 — will link this PR there.

egyptiangio (commented):
## Review

**Overview**: Reworks Stream Priority from first-match-wins into a two-class model — additive `score` rules (signed points, summed) ranked within hard `priority` bands, collapsed into one sortable int (`band * 1_000_000 - clamped_score`). Legacy configs stay byte-identical (stride only engages when a score rule exists). Includes settings-registry parsing, API validation, a two-section UI, an upgraded explainer popover with staleness via a recomputed `expected_priority`, docs, and 30+ tests across service/model/API layers.

### Strengths
- The design is right: one collapsed sort key means zero changes to every downstream consumer — I verified `generation.py` writes it to a SQLite `INTEGER` (64-bit, fine at ~10⁹ scale) and Dispatcharr receives **ordered ID lists**, never the raw number, so nothing downstream cares about magnitude.
- Byte-identical migration is genuinely byte-identical: `_has_score_rules` gating means pure-priority configs never touch the stride. The 5ag.3 close reason ("no migration needed, JSON defaults at parse time") checks out — `_parse_ordering_rules` + `update_stream_ordering_rules` + the Pydantic model all default `mode='priority'`, and the deliberate choice to keep the **storage** default legacy-safe while the **UI** defaults new rules to score is exactly the right split (and documented at each layer).
- `_SCORE_CLAMP = BAND_STRIDE // 2 - 1` with API bounds ±100k makes band-crossing arithmetically impossible; there's a test proving it.
- Nice catch on `isinstance(points, bool)` — bool-as-int would have slipped through.
- The `expected_priority` approach fixes a real pre-existing flaw in the staleness check (the old `winner.priority !== priority` comparison couldn't work once priorities are collapsed).

### 🐛 One bug to fix before merge — band-1 decode in the UI

`ManagedChannelsTable.tsx` `decodePriority`:

```ts
if (priority < BAND_STRIDE) return { band: priority, score: 0, scored: false }
```

A stream in **band 1 with a positive score** collapses to `1 * 1_000_000 - score` — e.g. `999_975` for +25 — which is **< BAND_STRIDE**, so it takes the "plain legacy value" branch and the cell renders **`999975`** instead of `1 +25`. Band 1 is the most common hard band (your own docs example uses priority 1), so most-preferred scored streams hit this.

Since the clamp guarantees collapsed values are ≥ `500_001` and plain legacy values are ≤ `999` (`NO_MATCH_PRIORITY`), the fix is a threshold change:

```ts
if (priority <= 999) return { band: priority, score: 0, scored: false }  // legacy/pure-priority
const band = Math.round(priority / BAND_STRIDE)  // safe: |score| < STRIDE/2
```

Worth a companion test if you add any frontend coverage, or at minimum a round-trip assertion in `test_stream_ordering.py` documenting that band-1-positive-score collapses below the stride (that's the invariant the UI decode must respect).

### Minor (non-blocking)
- `docs_version: "2.10.0"` in `stream-priority.md` guesses the next release; current is 2.9.1 and the release cadence is batched — consider leaving it at the version that actually ships it.
- `handleAddScoreRule` seeds `type: "regex"` with an empty value; if the user saves immediately, the backend silently drops the empty-value rule. A disabled-save-until-valid or a toast would avoid a "my rule vanished" report.
- `BAND_STRIDE` is duplicated in the frontend (drift risk with the backend constant). The comment cross-references it, which is probably fine — an alternative is having the channels endpoint return `band`/`score` explicitly since it computes them anyway, making the UI decode (and this bug class) disappear.

### Verdict
Solid work — approve once the band-1 decode is fixed. Everything else checks out: correctness of the collapse math, lossless migration, API validation, test coverage across all three layers, and docs that actually explain the model.

lpukatch:
Made the changes in question

egyptiangio (approved):
Band-1 decode fix verified — splitting on `NO_MATCH_PRIORITY` with the invariant documented at the constant, plus the backend round-trip test pinning `999_975` is exactly the guard this needed. Thanks for picking up the `docs_version` nit too. Merging.

PR #350, opened 2026-07-10
fix(matching): racing streams misclassified as TEAM_VS_TEAM; stream-name racing fallback (#349)

### LINKED ISSUE #349: Racing streams with single-word series names misclassify as TEAM_VS_TEAM; no racing fallback on stream-name path
Two related gaps keep racing streams from ever reaching the racing matcher after failing team-vs-team classification:

**1. Classifier separator guard swallows single-word series names**

`_classify_racing_event()` guards against team-sport streams leaked into racing groups by treating any single-word text before an `at`/`@` separator as a team abbreviation (like `"SD at BAL"`). But real series names are often single words — so in a racing group (`league_event_type="event"`):

- `NASCAR @ Daytona` → `TEAM_VS_TEAM` (team1="NASCAR", team2="Daytona") ❌
- `F1 at Monaco` → `TEAM_VS_TEAM` ❌
- `IndyCar at Indy`, `MotoGP at Mugello` → `TEAM_VS_TEAM` ❌
- `NASCAR Cup at Daytona` → `RACING_EVENT` ✅ (multi-word left side)

The misclassified streams route to TeamMatcher, find no team named "NASCAR", and end NO_MATCH.

**2. Racing fallback exists only on the EPG-title path**

The mixed-group racing fallback added in `37499266` (#263) — re-classify with `league_event_type="event"` + require `has_racing_text_evidence()` — lives only in `_match_via_epg`. The plain stream-name path (`_match_single` → `_route_to_outcomes`) classifies once and routes once, so in mixed groups (dominant type `team_vs_team`) racing streams have no second chance.

**Fix**

- Classifier: don't set `has_team_pattern` when the single-word left side is itself racing text evidence (`NASCAR`, `F1`, `MotoGP`, …).
- Matcher: extract the EPG racing fallback into a shared helper and call it from `_match_single` when the primary outcome is unmatched and the group includes an `event`-type league. Keep the `has_racing_text_evidence` guard.

tomwinterrose:
Fixes #349.

**1. Classifier separator guard** (`_classify_racing_event`): a single-word left part before `at`/`@` was always treated as a team abbreviation (`"SD at BAL"`), so `NASCAR @ Daytona`, `F1 at Monaco`, `IndyCar at Indy`, `MotoGP at Mugello` classified `TEAM_VS_TEAM` even in racing groups and died in TeamMatcher. The guard now exempts single-word left parts that are themselves racing series names (checked via `has_racing_text_evidence` on the raw left text — `"F1"` is under the 3-char team-extraction minimum, so `sep_team1` can't be used). `SD at BAL` still falls through to team matching.

**2. Stream-name racing fallback**: the mixed-group racing fallback from `37499266` (#263) existed only on the EPG-title path. It's now extracted into a shared `_try_racing_fallback` helper called from both:
- `_match_via_epg` (unchanged behavior, now via the helper)
- `_match_single`, when the primary route yields no match, and at the TEAM_ONLY/team-streams-disabled gate (mirroring the EPG path's bypass)

The stream-name call sites are gated on `name_match_enabled` (racing-by-name is a Stream Name matching type; TEAM_ONLY routing runs even with name matching off and must not leak into racing). The `has_racing_text_evidence` guard is preserved.

**Tests**: classification regression for single-word series (`NASCAR @ Daytona` et al.), plus four `_match_single` fallback tests (mixed-group match, text-evidence guard, name_match gating, TEAM_ONLY-gate bypass). Full suite: 1455 passed; ruff clean; frontend build clean.


egyptiangio:
Reviewed — this is exactly right on both counts. The single-word-series guard fix checks the raw left text (correctly, since `F1` is under the extraction minimum), and extracting the mixed-group fallback into `_try_racing_fallback` shared by both paths preserves the text-evidence guard that keeps arbitrary EPG programmes (the Brimstone→Silverstone case) out of date-coverage matching. Gating the stream-name call sites on `name_match_enabled` matches the per-source matching-type design.

I merged current dev into the branch (only `.beads/*.jsonl` bookkeeping conflicted) and re-ran the full suite: 1551 passed, ruff clean. Merging once CI re-confirms. Thanks @tomwinterrose!

Issue #392, opened 2026-07-11
perf: EPG generation loop speedups — parallel fetches, negative caching, sync diff-loop fixes


lpukatch:
## Problem

Hourly full EPG generation runs took ~280–540s, with the hidden channel-source group ("Dispatcharr Channels") alone consuming 100–230s (~45% of the run). Investigation with per-phase instrumentation and cProfile found:

1. **EPG program index fetch was serial** — one Dispatcharr round-trip per resolved tvg_id (~92 sequential calls), and mirror streams resolving to the same EPG-source tvg_id each triggered their own fetch.
2. **EPG match plan recomputed per stream** — 726 channel-source streams share 89 guide channels, but the classify+route pass over each program timeline ran once per *stream* (~8× redundant work).
3. **Event enrichment was serial** — `refresh_event_status` per matched stream, one at a time (~160–190 `espn:summary` calls in sequence).
4. **Provider misses were never cached** — `get_team`/`get_team_stats` returning None refetched every run: a steady ~250–300 `espn:teams` calls per hour, forever.
5. **"All profiles" sync diff loop (bug)** — `_parse_profile_ids` filtered with `if x`, dropping the `0` ALL-profiles sentinel, so every channel's stored `[0]` read back as `[]` and re-PATCHed "profiles: all profiles" every run (2,907 history rows / 3h observed).
6. **Windowed-stream tug-of-war** — settings sync re-added time-windowed EPG streams that were correctly detached outside their attach/detach window, fighting the ordering-phase window sync every run.
7. **Per-group work repeated per run** — lifecycle service construction, external-occupied computation, scheduled deletions (161×), and channel reassignment (2× per group = ~322 full scans) all ran per group.

## Results (measured on live-mirror data, warm runs)

| Metric | Before | After |
|---|---|---|
| Full generation run | ~278–542s | ~244s |
| Channel-source group | 103–228s | **48s** |
| Provider HTTP calls/run | 746–912 | ~417 |
| Dispatcharr PATCHes (channel-source group) | 1,019 | ~200 (legit window flips) |
| Profile-sync history churn | ~650 rows/run | 0 |

Also adds per-phase wall-time instrumentation (`extra_metrics.phase_timings` on full runs and per-group runs) so future regressions are visible in run stats.

## Follow-up candidates (not in this change)

- Channel-source candidate fetch pulls all ~24k Dispatcharr streams (~12–15s)
- Dispatcharr EPG refresh wait (~16–23s, external)
[...]

PR #393, opened 2026-07-11
perf: EPG generation loop speedups — parallel fetches, negative caching, sync diff-loop fixes (#392)

lpukatch:
Closes-when-released #392 · bead `teamarrv2-44ey`

## What

Seven targeted fixes to the full-generation hot path, found via new per-phase instrumentation + cProfile against live-mirror data:

**Round 1 — fetch/call volume**
- `EPGProgramIndex.build`: dedupe fetches by resolved EPG-source tvg_id (mirror streams shared one guide channel but each fetched it) and fan out over a thread pool (`EPG_INDEX_FETCH_WORKERS`, default 10; documented in configuration.md). 92 serial round-trips → 3.2s parallel.
- `_enrich_matched_events`: refresh each unique event once, in parallel (was serial per matched stream; `refresh_event_status` is already single-flight + coalesced).
- `get_team` / `get_team_stats`: negative-cache provider misses (`CACHE_TTL_NEGATIVE`, 4h) — kills a steady ~250–300 repeated `espn:teams` calls per hourly run.
- `EventGroupProcessor`: one lifecycle service per run (the Dispatcharr channel cache is class-level; per-group construction bought nothing), scheduled deletions once per run (was 161×), pre-reassign once per run, post-reassign only when a group changed the channel set (was 2× per group ≈ 322 full scans/run).

**Round 2 — per-run sync churn (profiler findings)**
- `StreamMatcher`: memoize the EPG-title match plan per tvg_id — 726 channel-source streams share 89 guide channels; the classify+route pass ran once per stream (~8× redundant). Now computed once per tvg_id, per-stream results materialized from the shared plan.
- **Bug:** `_parse_profile_ids` filtered with `if x`, dropping the `0` ALL-profiles sentinel — every channel's stored `[0]` read back as `[]` and re-PATCHed "profiles: all profiles" every run (2,907 history rows / 3h on live data). Fixed the parse; also compare Dispatcharr's expanded concrete profile list against the profile catalog when effective is the sentinel (reads never return `[0]`); an empty Dispatcharr list stays drift (self-healing preserved, covered by existing test).
- `_sync_channel_settings`: skip the stream drift-add for time-windowed (EPG-matched) streams — an out-of-window stream is correctly absent from Dispatcharr; the re-add fought the ordering-phase window sync every run.

**Instrumentation**
- Per-phase wall times on full runs and per-group runs (`extra_metrics.phase_timings`), fetch timing on the `[EPG-INDEX]` log line — regressions now show in run stats instead of requiring log archaeology.

## Measured (live-mirror data, warm hourly runs)

| Metric | Before | After |
|---|---|---|
| Full generation run | ~278–542s | **~244s** |
| Channel-source group ("Dispatcharr Channels", 726 streams) | 103–228s | **48s** (fetch 15 / match 15 / enrich 0.2 / channels 15) |
| Provider HTTP calls per run | 746–912 | **~417** |
| Dispatcharr PATCHes (channel-source group) | 1,019 | ~200 (remaining are genuine attach/detach window flips) |
| Profile-sync history churn | ~650 rows/run | **0** |

## Tests

- 1,621 passing (`pytest tests/ -v`), ruff clean, frontend build green.
- New: EPG-index fetch-dedupe test; negative-cache tests for team/stats (mirroring `test_event_negative_cache.py`); regression tests for the `[0]` sentinel parse and no-repatch-when-expanded behavior.

https://claude.ai/code/session_018bqfTnGE8jY8ojmqKvYM7G

### DIFF
--- a/docs/reference/deployment/configuration.md
+++ b/docs/reference/deployment/configuration.md
@@ -19,6 +19,7 @@ Teamarr is configured via environment variables in your `docker-compose.yml` fil
 | `LOG_FORMAT` | `text` | Log format: `text` or `json` (for log aggregation systems like ELK, Loki, Splunk) |
 | `LOG_DIR` | auto-detected | Override log directory path. See [Log Directory Detection](#log-directory-detection). |
 | `SKIP_CACHE_REFRESH` | `false` | Skip team/league cache refresh on startup. Set to `true`, `yes`, or `1`. Useful for faster restarts during development. |
+| `EPG_INDEX_FETCH_WORKERS` | `10` | Parallel workers for fetching Dispatcharr EPG programs during EPG matching. Lower it if your Dispatcharr instance struggles with concurrent requests. |
 
 ## ESPN API Settings
 
--- a/teamarr/consumers/event_group_processor/matching.py
+++ b/teamarr/consumers/event_group_processor/matching.py
@@ -645,26 +645,48 @@ class StreamMatching:
         if not matched_streams:
             return matched_streams
 
+        # Refresh each unique event once, in parallel. The service coalesces
+        # repeated refreshes of the same event within a run (single-flight in
+        # get_event), so this is safe across threads — the dedupe here just
+        # avoids queueing redundant no-op calls.
+        from concurrent.futures import ThreadPoolExecutor
+
+        unique_events = {}
+        for match in matched_streams:
+            event = match.get("event")
+            if event:
+                unique_events.setdefault((event.league, event.id), event)
+
+        refreshed_by_key = {}
+        if unique_events:
+            workers = min(10, len(unique_events))
+            with ThreadPoolExecutor(max_workers=workers) as executor:
+                for key, refreshed in zip(
+                    unique_events.keys(),
+                    executor.map(
+                        self._service.refresh_event_status, unique_events.values()
+                    ),
+                    strict=True,
+                ):
+                    old = unique_events[key]
+                    old_status = old.status.state if old.status else "N/A"
+                    new_status = refreshed.status.state if refreshed.status else "N/A"
+                    if old_status != new_status:
+                        logger.debug(
+                            "[ENRICH] event=%s status changed: %s → %s",
+                            old.id,
+                            old_status,
+                            new_status,
+                        )
+                    refreshed_by_key[key] = refreshed
+
         enriched = []
         for match in matched_streams:
             event = match.get("event")
             if event:
-                old_status = event.status.state if event.status else "N/A"
-                # Refresh event status from provider. The service coalesces
-                # repeated refreshes of the same event within a run, so an event
-                # matched to many channels triggers a single provider fetch.
-                refreshed = self._service.refresh_event_status(event)
-                new_status = refreshed.status.state if refreshed.status else "N/A"
-                if old_status != new_status:
-                    logger.debug(
-                        "[ENRICH] event=%s status changed: %s → %s",
-                        event.id,
-                        old_status,
-                        new_status,
-                    )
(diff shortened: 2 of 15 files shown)


egyptiangio:
Triage: this is excellent work — seven targeted fixes with per-phase instrumentation and measured before/after on live-mirror data, plus a real bug fix (the `[0]` ALL-profiles sentinel being dropped by `_parse_profile_ids`, causing ~650 no-op PATCH rows/run). The profile-sync change touches the self-healing path in `lifecycle/service.py`, so review will focus there and on the windowed-stream drift-add skip. Expect a full review within a day; targeting the v2.10.x cycle. Tracking bead `teamarrv2-w5tx` (#392).

egyptiangio (approved):
## Review

Went through all seven fixes with the surrounding code. This is a model perf PR — instrumented first, each fix targeted and separately explainable, measured on live-mirror data. **Approving**; a few non-blocking notes below.

### Verified against the codebase

- **Profile sentinel fix** — the `_parse_profile_ids` fix is right, and the syncer's sentinel comparison is sound: superset-of-catalog check handles Dispatcharr's read-expansion, an *empty* Dispatcharr list correctly stays drift (self-healing preserved — the regression test covers exactly the case I was worried about), and an unavailable catalog conservatively declines to churn. I also checked the settings registry's separate `_parse_profile_ids` in `database/settings/registry.py` for a sibling bug — it has intentionally different semantics (`None` = all profiles) and `"[0]"` round-trips fine there, so no companion fix needed.
- **Enrichment dedupe is lossless** — matched streams share Event instances per event id (the matcher emits from the shared event list; UFC segment info lives on the match dict, not the event), and `refresh_event_status` overlays game-state onto its input, so collapsing to one refresh per `(league, id)` returns the identical object the serial path produced. Exception behavior is also parity: `executor.map` propagates like the old serial loop did.
- **Per-run consolidation can't leak across runs** — every module-level entry point (`process_single_event_group`, `process_all_event_groups`, `preview_event_group`) constructs a fresh processor, so the `_deletions_processed`/`_initial_reassign_done` flags and the shared lifecycle service are genuinely scoped to one run. Factory arg order matches the positional call.
- **Match-plan memo** — the plan stores `(outcome, classified)`, which are stream-independent; per-stream identity is re-applied at `_outcome_to_result`. First-carrier `stream_id` leaking into sub-matcher cache keys is harmless.
- **Windowed drift-add skip** — both `_sync_channel_settings` call sites pass the new flag, and window membership is owned by the ordering-phase sync, so the skip can't strand an in-window stream.
- Function-level DB imports stayed function-level (that's a deliberate test seam in the consumers — good instinct not to hoist).

### Non-blocking notes

1. **Post-reassign gate** (`combined_result.created or combined_result.deleted`): an in-place channel *update* that changes its numbering bucket wouldn't trigger the pass. I believe that can't happen today (event changes go through delete/create, and full runs are backstopped by the generation-level reassign), but a one-line comment stating the assumption would help the next reader.
2. `FETCH_WORKERS = int(os.environ.get(...))` crashes at import on a malformed value. Consistent with how the ESPN_* envs are read, so fine — just noting it's now two places that share the pattern if we ever harden it.
3. **Negative cache**: a transient provider outage (caught-and-returns-None paths) now poisons team/stats lookups for 4h. Same tradeoff we already accepted for `_EVENT_NOT_FOUND`, and the TTL is short enough — just flagging that the failure mode now extends to template team/stats vars.

### Tests & CI

New coverage is exactly where it should be (fetch-dedupe, sentinel parse + no-repatch-when-expanded, negative-TTL). All checks green including CodeQL. Merging after this review lands.

PR #394, opened 2026-07-11
Fix WEC never matching + scope racing session fan-out by stream-name session labels

tomwinterrose:
## Summary

Two related racing-matching fixes, found while running WEC and NASCAR event groups in production:

### 1. WEC streams never match (`classifier.py`)

`RACING_TEXT_EVIDENCE` has no signature for the FIA World Endurance Championship, so genuine WEC streams (`FIA WEC | ROUND 5 | 6 HOURS OF SAO PAULO | ...`, `WEC | Qualifying: ...`) can never pass the `has_racing_text_evidence` gate on the mixed-group racing fallback in `matcher.py`. The result is that WEC events systematically produce **zero** stream matches regardless of what the provider carries — `wec` is in `_RACING_LEAGUE_HINTS` and the schema, but no text can ever prove it. Fix: add `wec|world\s+endurance` to the regex.

### 2. Racing session fan-out ignores session labels in stream names (`racing_segments.py`)

`expand_racing_segments` fans every matched racing stream out into a channel entry for *every* session in the event. In practice that means a dedicated "Free Practice 3" feed also shows up as a source on the Qualifying and Race channels, and "NASCAR Cup Series Qualifying" streams land on the Race channel.

This PR scopes the fan-out when the stream's own name clearly names one session:

- **Label-is-session-type** convention (STAN/tsdb style): `AU (STAN 36) | Free Practice 3: 6 Hours of Sao Paulo WEC 2026 (...)` → fp3 only.
- **Label-ends-in-session-type** convention (TSN+ style): `CA (TSN+ 021) | NASCAR Cup Series Qualifying: Quaker State 400 (...)` → qualifying only. Trailing-keyword matching is end-anchored, so branding like "Race Week Live" doesn't false-positive.
- A bare **"Qualifying"** label also covers WEC **Hyperpole** sessions (Hyperpole is a qualifying shootout; providers rarely label it distinctly). The reverse is not true — an explicit "Hyperpole" stream stays scoped to hyperpole sessions.
- Detection is deliberately conservative: a name with no recognizable session label (generic linear channels like `TNT Sports 4 HD`) keeps the previous full fan-out, and a label that matches no session in *this* event's session list falls back to the full fan-out rather than silently producing nothing.

It also **excludes practice sessions from the generic-stream fan-out** (still produced when a stream specifically names one): providers frequently don't carry dedicated practice feeds at all, so a whole-weekend stream landing on a Practice channel tends to be dead air until qualifying starts — better no channel than that. If this is too opinionated a default for teamarr, this part is cleanly separable; happy to drop or gate it behind a setting.

The session-label regexes are defined locally rather than imported from `providers.tsdb.racing`: that parser's `_FREE_PRACTICE_RE` requires a digit and the literal "free practice"/"fp" forms, so it would miss stream labels like "Practice 2".

### 3. PLACEHOLDER streams never reach the racing fallback (`matcher.py`)

A racing stream carrying a timestamp but no team separator (`US (Peacock 031) | IMSA CTMP Grand Prix (2026-07-12 14:00:00)`) classifies PLACEHOLDER in a team-dominant group — the extracted date/time disqualifies `_classify_team_only`, and nothing else fits. `_match_single` then returned `"unclassifiable"` at step 2, **before** the #349 racing fallback was consulted, so a stream the fallback would happily match (series name present, single covering event, fuzzy 66 vs. threshold 50) was dropped instead. Fixed by routing the PLACEHOLDER branch through `_try_racing_fallback` first, under the same `name_match_enabled` gate as the TEAM_ONLY path; timestamped streams *without* racing text evidence still short-circuit as unclassifiable.

### 4. Cross-series date-binding: streams for unconfigured series match whatever races that weekend (`matcher.py`, `classifier.py`)

Found live right after fix 3 landed: two "MotoGP - Grand Prix of Germany" streams direct-matched IMSA's "Chevrolet Grand Prix". MotoGP isn't a configured league, so the IMSA event was the only racing event covering the date, and the shared "Grand Prix" tokens cleared the single-event sanity bar — the text-evidence gate can't reject it because MotoGP *is* racing text. Fixed with per-series signature patterns (`detect_racing_series_leagues`, a scoping subset of `RACING_TEXT_EVIDENCE`): when a stream explicitly names a series, `_match_racing_event` only tries that series' league(s), and fails outright when none are configured. Generic racing text (a bare "Monaco Grand Prix") names no series and stays unscoped, preserving existing behavior.

## Testing

- 35 new tests across the four fixes covering label isolation, both label conventions, hyperpole/qualifying coverage, practice exclusion, the no-matching-session fallback, and the WEC text-evidence regression.
- Full suite: **1665 passed** on this branch.
- Both fixes verified live against a production Dispatcharr + teamarr:dev install during a WEC race weekend (6 Hours of São Paulo) and a NASCAR triple-header weekend: WEC streams matched only after fix 1, and per-session streams stopped attaching to the wrong session channels with fix 2.

Closes-when-released #399

### DIFF
--- a/teamarr/consumers/matching/classifier.py
+++ b/teamarr/consumers/matching/classifier.py
@@ -1116,6 +1116,13 @@ def has_racing_text_evidence(text: str) -> bool:
 # naming them blocks cross-series binding entirely.
 _RACING_SERIES_SIGNATURES: "tuple[tuple[Pattern[str], tuple[str, ...]], ...]" = (
     (re.compile(r"\b(?:formula\s*1|f1)\b", re.IGNORECASE), ("f1",)),
+    # F2/F3/Formula E run as support series on F1 weekends at the SAME venue,
+    # so an unscoped "Formula 2 - Monaco - Sprint Race" stream shares venue
+    # tokens with the one configured F1 event covering that date — the exact
+    # cross-series shape this table exists to block, at a much higher
+    # collision rate (every F1 weekend). No corresponding league exists in
+    # schema.sql, so like supercross below this maps to a blocking sentinel.
+    (re.compile(r"\b(?:formula\s*[23e]|f[23])\b", re.IGNORECASE), ("formula-feeder",)),
     (
         re.compile(r"\bnascar\b", re.IGNORECASE),
         ("nascar-cup", "nascar-xfinity", "nascar-truck"),
--- a/teamarr/consumers/matching/classifier.py
+++ b/teamarr/consumers/matching/classifier.py
@@ -1107,6 +1107,45 @@ def has_racing_text_evidence(text: str) -> bool:
     return bool(text) and RACING_TEXT_EVIDENCE.search(text) is not None
 
 
+# Per-series text signatures, for scoping a racing stream to the league(s)
+# whose series it explicitly names. Values are the league codes the series
+# can belong to (NASCAR is an umbrella over three leagues). Deliberately a
+# subset of RACING_TEXT_EVIDENCE: generic evidence like "grand prix" names
+# racing but not a series, and stays unscoped. Series with no corresponding
+# league (supercross/motocross) map to a code no group will ever include, so
+# naming them blocks cross-series binding entirely.
+_RACING_SERIES_SIGNATURES: "tuple[tuple[Pattern[str], tuple[str, ...]], ...]" = (
+    (re.compile(r"\b(?:formula\s*1|f1)\b", re.IGNORECASE), ("f1",)),
+    (
+        re.compile(r"\bnascar\b", re.IGNORECASE),
+        ("nascar-cup", "nascar-xfinity", "nascar-truck"),
+    ),
+    (re.compile(r"\b(?:indycar|indy\s*(?:500|nxt))\b", re.IGNORECASE), ("indycar",)),
+    (re.compile(r"\b(?:motogp|moto\s*[23])\b", re.IGNORECASE), ("motogp",)),
+    (re.compile(r"\bimsa\b", re.IGNORECASE), ("imsa",)),
+    (re.compile(r"\b(?:wec|world\s+endurance)\b", re.IGNORECASE), ("wec",)),
+    (re.compile(r"\b(?:supercross|motocross)\b", re.IGNORECASE), ("supercross",)),
+)
+
+
+def detect_racing_series_leagues(text: str) -> "tuple[str, ...] | None":
+    """League codes for the racing series the text explicitly names.
+
+    Returns None when no specific series is named (e.g. a bare "Monaco
+    Grand Prix") — callers must treat that as "unscoped", not "not racing".
+    A stream naming a specific series must only match events from that
+    series' league(s): a MotoGP stream must not date-bind to the one IMSA
+    event covering the weekend just because motogp isn't configured.
+    """
+    if not text:
+        return None
+    hits: list[str] = []
+    for pattern, leagues in _RACING_SERIES_SIGNATURES:
+        if pattern.search(text):
+            hits.extend(leagues)
+    return tuple(hits) or None
+
+
 # Known tennis league codes — mirror of the sport='tennis' leagues in
 # schema.sql (this module is pure text classification, no DB access).
 _TENNIS_LEAGUE_HINTS: frozenset[str] = frozenset({"atp", "wta"})
--- a/teamarr/consumers/matching/classifier.py
+++ b/teamarr/consumers/matching/classifier.py
@@ -1096,7 +1096,7 @@ def is_racing(
 RACING_TEXT_EVIDENCE: Pattern[str] = re.compile(
     r"\b(?:"
     r"formula\s*[123e]|f1|nascar|indycar|indy\s*(?:500|nxt)|"
-    r"motogp|moto\s*[23]|grand\s+prix|imsa|supercross|motocross"
+    r"motogp|moto\s*[23]|grand\s+prix|imsa|supercross|motocross|wec|world\s+endurance"
     r")\b",
     re.IGNORECASE,
 )
--- a/teamarr/consumers/matching/matcher.py
+++ b/teamarr/consumers/matching/matcher.py
@@ -597,6 +597,22 @@ class StreamMatcher:
         # This handles streams that passed filtering but still can't be classified
         # (e.g., no separator found, no custom regex match).
         if classified.category == StreamCategory.PLACEHOLDER:
+            # A racing stream with a timestamp but no separator ("US (Peacock
+            # 031) | IMSA CTMP Grand Prix (2026-07-12 14:00:00)") classifies
+            # PLACEHOLDER — a date/time was extracted, so _classify_team_only
+            # refuses it and nothing else fits. Give the racing fallback a
+            # chance before writing it off as unclassifiable (same gate as
+            # the TEAM_ONLY path below).
+            if self._name_match_enabled:
+                fallback = self._try_racing_fallback(stream_name, stream_id, target_date)
+                if fallback is not None:
+                    outcome, racing_classified = fallback
+                    return [self._outcome_to_result(
+                        outcome=outcome,
+                        stream_id=stream_id,
+                        stream_name=stream_name,
+                        classified=racing_classified,
+                    )]
             return [MatchedStreamResult(
                 stream_name=stream_name,
                 stream_id=stream_id,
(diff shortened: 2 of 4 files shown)


egyptiangio:
Triage: both fixes look well-scoped and the live verification during an actual WEC/NASCAR weekend is exactly the evidence we want. The WEC text-evidence gap is a clear bug — that part is a fast merge. The session-label fan-out scoping is a real behavior change we'll review more carefully, and the practice-session exclusion from generic fan-out is the one opinionated default to decide (leaning toward taking it as proposed given the dead-air rationale, but we may gate it behind a group setting — will confirm in review). Could you open an issue for this so it tracks through the release checklist? If not, no problem — I'll file one when I review. Expect a full review within a day or two, targeting the v2.10.x cycle. Thanks as always @tomwinterrose.

egyptiangio (changes requested):
## Review

Filed #399 to track this (bead `teamarrv2-5ckg`) — please add `Closes-when-released #399` to the PR body. The WEC fix is verified correct, the scoping architecture is exactly right (conservative fallbacks, asymmetric hyperpole↔qualifying coverage is well-reasoned), and the test coverage is thorough. One real gap to fix before merge, one decision confirmed, one comment nit.

### Requested change: unnumbered trailing "Practice" inverts for NASCAR

`_TRAILING_FP_RE` requires a digit (`(\d)`), but the NASCAR provider emits an **unnumbered `"practice"` session code** (`teamarr/providers/nascar/provider.py:66` — `return "practice", event_name or "Practice"`). So a TSN+-convention stream — your own evidence convention —

```
CA (TSN+ 021) | NASCAR Cup Series Practice: Quaker State 400 (...)
```

gets no session hint (the anchored `_FREE_PRACTICE_RE` needs the label to BE "Practice"; the trailing fallback needs a digit), falls into the generic fan-out, and the practice exclusion then drops the practice session — the dedicated practice feed lands on the Qualifying and Race channels and **not** on Practice. Exactly inverted.

Suggested fix: make the digit optional — `(?:free\s*practice|practice|fp)\s*(\d?)\s*$` — and map an empty group to the `"practice"` category (mirroring what `_FREE_PRACTICE_RE` already does for a bare "Practice" label). A label *ending* in "practice" with no digit is still clearly a practice feed; false-positive surface looks negligible. Please add a test for the unnumbered trailing form against a `"practice"`-coded session.

### Decision: practice exclusion — taking it as proposed

Confirmed: we'll take the practice-session exclusion from the generic fan-out **as the default, ungated**. Rationale: dedicated practice feeds still produce practice channels (especially once the fix above lands), the dead-air failure mode is visible and ugly, and you verified live. If Sky-F1-style whole-weekend-linear users push back, we'll gate it behind a group setting then — noted for the changelog.

### Nit

The comment above `_TRAILING_FP_RE` says "'Pre-Race Show' still counts as naming the race" — it doesn't (it ends in "Show", so `\brace\s*$` won't match; "Pre-Race" alone would). Worth rewording so the comment matches the regex.

Everything else checked out against the codebase: `import re` already present, `wec` already in `_RACING_LEAGUE_HINTS`/schema so the evidence regex was genuinely the only missing piece, and the no-matching-session fallback keeping FULL fan-out (practice included) is a sensible defensive asymmetry. CI is green. Fix the trailing-practice gap and this merges.

tomwinterrose:
Thanks for the careful review — the trailing-practice inversion was a genuinely good catch, and it reproduces exactly as you described (I confirmed our NASCAR provider path emits the unnumbered `"practice"` code).

Addressed in 2297d39:

- **Trailing practice fix**: digit is now optional with an empty capture mapping to the `"practice"` category, mirroring `_FREE_PRACTICE_RE`'s bare-"Practice" handling. I also added a leading `\b` — with the digit no longer anchoring the match, a word-suffix like "Malpractice" would otherwise trail-match; there's a negative test for that alongside the two you asked for (label-level category test + an `expand_racing_segments` test against a `"practice"`-coded session, asserting the feed lands on practice only).
- **Comment nit**: reworded — "Pre-Race" is the example that trail-matches race; "Pre-Race Show" and "Racecourse Network" are the ones that don't.
- **Tracking**: added `Closes-when-released #399` to the PR body.

Good to hear the practice exclusion lands ungated. Full suite is at 1657 passed on the branch.

tomwinterrose:
One more commit (found live overnight): the PLACEHOLDER racing fallback exposed a pre-existing cross-series binding gap — two "MotoGP - Grand Prix of Germany" streams direct-matched IMSA's "Chevrolet Grand Prix" (motogp isn't configured, the IMSA event was the only racing event covering the date, and the shared "Grand Prix" tokens cleared the single-event sanity bar; text evidence can't help since MotoGP *is* racing text).

Fix: per-series signature patterns — when a stream explicitly names a series, `_match_racing_event` only tries that series' league(s) and fails outright when none are configured. Generic racing text (a bare "Monaco Grand Prix") names no series and stays unscoped, preserving existing behavior. Suite now at 1665 passed.

egyptiangio (changes requested):
## Re-review

Everything from the last round is addressed — the unnumbered trailing "Practice" fix is exactly right (empty digit group → `"practice"` category, mirroring `_FREE_PRACTICE_RE`), both new tests cover the inversion scenario, the comment is reworded, and `Closes-when-released #399` is in the body. Thanks for the fast turnaround.

The two new fixes (3 & 4) also check out:

- **Fix 3 (PLACEHOLDER → racing fallback):** verified `_try_racing_fallback` still gates on `has_racing_text_evidence` + an actual match, so dated filler streams stay unclassifiable — and you have the negative test asserting the racing matcher is never called for them. Same `name_match_enabled` gate as the TEAM_ONLY path. Correct.
- **Fix 4 (per-series scoping):** the architecture is right — signatures as a deliberate subset of `RACING_TEXT_EVIDENCE`, generic "grand prix" unscoped, NASCAR umbrella, supercross/motocross as a blocking sentinel.

### One more round: Formula 2/3/E repeat the exact failure mode fix 4 targets

`RACING_TEXT_EVIDENCE` matches `formula\s*[123e]`, but `_RACING_SERIES_SIGNATURES` only scopes `formula\s*1|f1`. So a "Formula 2 – Monaco – Sprint Race" stream carries racing evidence, names no scopeable series, and stays **unscoped** — and F2/F3 run as support series *on F1 weekends at the same venue*, so it shares venue tokens with the one configured F1 event covering that date. That's the identical shape to your "MotoGP direct-matched IMSA's Chevrolet Grand Prix" case, just with a much higher collision rate (every F1 weekend, not the occasional shared one). There are no f2/f3/formula-e leagues in `schema.sql`, so a blocking entry mirroring your supercross pattern should do it:

```python
(re.compile(r"\b(?:formula\s*[23e]|f[23])\b", re.IGNORECASE), ("formula-feeder",)),
```

…plus a test (an "F2 Sprint Race" / "Formula 2" stream must not bind to a configured F1 event). Ordering note: make sure the F1 pattern can't also fire on the same text — `formula\s*1|f1` won't match "Formula 2", so a plain additional tuple is fine.

Have a look and see if you agree — if so, that's the only thing between this and merge. (Moto2/Moto3 → `motogp` I'm taking as-is: same event weekend, one league, defensible.)

CI is green across the board. One more round and this ships.

tomwinterrose:
Agreed on all points — F2/F3/Formula E is the same shape as the MotoGP/IMSA case with a worse collision profile (shared venue tokens on every F1 weekend, not just coincidentally-shared dates), and the blocking sentinel is the right mechanism given there's no feeder league in schema.sql.

Addressed in 6a13302:

- Added `(?:formula\s*[23e]|f[23])` → `("formula-feeder",)` to `_RACING_SERIES_SIGNATURES`, positioned right after the F1 entry with a comment explaining the support-series collision rationale.
- Ordering confirmed as you noted: `formula\s*1|f1` can't fire on "Formula 2"/"F2", and there's an explicit assertion for that in the new detect test (F2/F3/Formula E → sentinel; "F1 | Monaco Grand Prix" still → `f1`).
- Matcher-level regression test: `test_racing_series_scoping_blocks_f2_from_binding_to_f1` — an "F2 | Monaco | Sprint Race" stream in a group with only f1 configured never reaches the racing matcher and ends unmatched.

Suite at 1667 passed. (Also mirrored the commit onto #405's stacked branch so it stays collapse-clean.)

egyptiangio (approved):
## Round 3 — approved

The F2/F3/Formula E blocking sentinel is exactly as discussed: the pattern sits right after the F1 entry with a comment explaining the support-series collision profile, and `detect_racing_series_leagues` accumulates hits so a title naming both F1 and F2 scopes to the union rather than blocking F1 outright. Both new tests cover the right things — the detection triple (F2/F3/Formula E → `formula-feeder`, F1 untouched) and the integration assertion that the racing matcher is never invoked for an F2 stream in an F1-only group.

CI green on `6a13302`. Merging — thanks for three fast, high-quality rounds on this one. #399 gets `status: on-dev` and closes at the next release (v2.10.1 is imminent).

Issue #399, opened 2026-07-11
WEC streams never match + racing session fan-out ignores stream-name session labels


egyptiangio:
Tracking issue for PR #394 (@tomwinterrose).

1. **WEC streams never match**: `RACING_TEXT_EVIDENCE` has no WEC signature, so genuine WEC streams can never pass the `has_racing_text_evidence` gate on the mixed-group racing fallback — WEC event groups systematically produce zero stream matches.
2. **Racing session fan-out ignores session labels in stream names**: `expand_racing_segments` fans every matched racing stream into a channel entry for every session in the event, so a dedicated "Free Practice 3" feed also shows up as a source on Qualifying/Race channels.

Fix in PR #394, verified live during a WEC/NASCAR weekend.

egyptiangio:
Bead: `teamarrv2-5ckg` — implementation in PR #394.

egyptiangio:
Landed on dev in PR #394 (merge 42b7cfd). Closes at the next release.

egyptiangio:
Shipped in [v2.11.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.11.0).

PR #405, opened 2026-07-12
Anchor EPG racing matches to the programme's broadcast instant

tomwinterrose:
## Summary

EPG-driven racing matches bind by **date coverage alone**: any programme whose title carries racing text evidence and clears the fuzzy bar matches "the one race covering this date" regardless of what hour it airs.

Live false positive that motivated this: **"Grand Prix of Europe" — the animated kids' movie — airing at 03:15 on Disney Junior matched IMSA's "Chevrolet Grand Prix" (race start 18:00)** at fuzzy 75 via the shared "Grand Prix" tokens, on every hourly run. The text-evidence gate can't reject it ("grand prix" *is* racing evidence), series scoping (#394) can't either (the title names no series), and both fuzzy thresholds pass. The only thing that distinguishes a broadcast of the race from a coincidentally-titled programme is *when it airs* — which the racing matcher never looked at.

## Changes

1. **Anchor gate in `RacingMatcher`** (ported from the team/event-card EPG matching's bead-t5e anchor concept, which `_route_to_outcomes` already threads — the racing branch just dropped it): the EPG path passes the programme's start instant, and candidate events are further gated to those with a session actually airing within ±90 min of it (`_covers_instant`, using each session's real `[start, end]` from `get_session_times`). Applied to fresh matches **and cache hits**, so a stale cached bind can't outlive the gate.

2. **Text-evidence gate on the PRIMARY EPG classification path**: in a racing-dominant group `_get_dominant_event_type()` returns `"event"` directly, so the primary `classify_stream` call defaults arbitrary EPG titles (documentaries, movies) to `RACING_EVENT` — the fallback's existing gate never runs for those groups. The same check now applies before routing.

3. **The stream-NAME path is untouched by design**: all name-path call sites pass `anchor_dt=None`, and the gate only fires when an anchor is present — the whole race weekend remains a legitimate match target for named streams, so #394's placeholder fallback, series scoping, and session-label scoping behave identically (full suite re-run green to confirm).

## Testing

- 6 new tests: `_covers_instant` unit coverage (near-session pass, hours-away reject incl. the Disney Junior reproduction, multi-session weekends), the primary-path evidence gate (unrelated documentary rejected / genuine WEC title still matches), and an EPG-plumbing test asserting the programme instant reaches the racing route as `anchor_dt`.
- Full suite: **1708 passed** on this branch (rebased onto current `dev` post-#394/#407, including the `_compute_epg_plan` perf refactor).
- Live-verified on the same install as #394: the movie's 03:15 and 15:15 showings both reject (race window with tolerance is 16:30–22:15), while during-race and 30-min-pre-race instants accept; genuine Peacock IMSA streams unaffected.

Closes-when-released: happy to link an issue if you'd like one filed like #399.

### DIFF
--- a/teamarr/consumers/matching/matcher.py
+++ b/teamarr/consumers/matching/matcher.py
@@ -727,7 +727,9 @@ class StreamMatcher:
         if classified.category == StreamCategory.EVENT_CARD:
             return [self._match_event_card(classified, stream_id, target_date)]
         if classified.category == StreamCategory.RACING_EVENT:
-            return [self._match_racing_event(classified, stream_id, target_date)]
+            return [self._match_racing_event(
+                classified, stream_id, target_date, anchor_dt=anchor_dt
+            )]
         if classified.category == StreamCategory.TENNIS_MATCH:
             return self._match_tennis_event(classified, stream_id, target_date)
         if classified.category == StreamCategory.TEAM_ONLY:
@@ -830,6 +832,25 @@ class StreamMatcher:
                 )
                 continue
 
+            # Same text-evidence gate as the racing fallback, applied to the
+            # PRIMARY classification too: in a racing-dominant group
+            # _get_dominant_event_type() returns "event" directly, so this
+            # classify_stream call already defaults arbitrary EPG titles
+            # (documentaries, movies) to RACING_EVENT with no series name in
+            # the text — the fallback's gate never even runs for these groups
+            # since primary_outcomes already "succeeds". Without this check,
+            # any program on any linear channel resolved into a racing-only
+            # group can bind to "the one race happening this weekend" by date
+            # coverage alone.
+            if classified.category == StreamCategory.RACING_EVENT and not (
+                has_racing_text_evidence(epg_input)
+            ):
+                logger.debug(
+                    "[EPG_MATCH] racing programme skipped, no series name in text: %s",
+                    epg_input[:60],
+                )
+                continue
+
             # TEAM_ONLY gate: skip team routing when disabled, but allow the
             # racing fallback to run if racing leagues are present. A race title
             # like "F1 | Monaco Grand Prix" classifies TEAM_ONLY in a mixed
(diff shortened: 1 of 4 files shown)


egyptiangio:
#394 is merged, so this can come off draft — rebase onto `dev` and it should reduce to just `771f4e0`. Ping when it's up and I'll review.

Assessment from reading the commit ahead of the formal review — this looks right:

- Porting the bead-t5e anchor concept is the correct move; `_route_to_outcomes` already threads `anchor_dt` and the racing branch had simply dropped it.
- Gating against each session's real `[start, end]` ± 90 min (rather than event start) keeps practice/quali broadcasts matchable, and the no-sessions degradation to `event.start_time` is sensible.
- Applying the gate to cache hits closes the stale-bind hole, and `anchor_dt=None` on the name path keeps the #394 behaviors provably untouched.

I'll file a tracking issue on our side (same pattern as #399) once it's ready for review.

tomwinterrose:
@egyptiangio Rebased onto `dev` — reduced to the single commit (`4f3465b`, same content as `771f4e0`) and marked ready for review. Full suite on the rebased branch: 1708 passed. Happy to link the tracking issue in the description once it's filed.

egyptiangio:
Tracking issue filed: #410 — please add `Closes-when-released #410` to the PR body. Review incoming.

egyptiangio (approved):
## Review — approved

Verified beyond the diff:

- **Timezone consistency:** the new comparisons (`event.start_time - anchor_dt`, session `[start, end]` vs `anchor_dt`) mix the same two datetime families dev already subtracts on every EPG run (the t5e skew diagnostic at `_compute_epg_plan`'s tail does `ev_start - program.start_dt`), so awareness consistency is established by working production code.
- **Golf/tennis safety of the primary-path evidence gate:** `is_racing`'s dominant-type trigger is already disabled for non-racing event-sports via `event_league_sport`, and even a legacy-`None` classification previously fell through the racing league-search without matching — the gate changes the failure detail, not any capability.
- **`_covers_instant` degradation** (no sessions → `event.start_time` ± tolerance) and the cache-hit gate both check out; the cache gate is what keeps a pre-fix stale bind from surviving the upgrade.
- Ran `tests/matching/test_racing.py` + `tests/epg/` on this branch merged with current dev (post-#407/#409): 193 passed.

Replay/re-air exclusion is intended behavior — live-event channel semantics are the app-wide norm.

CI green. Merging. #410 gets `status: on-dev` and closes at the next release.

PR #407, opened 2026-07-12
fix: make channel-source group EPG-only (#406)

### LINKED ISSUE #406: Channel-source group runs Team matching despite being labeled 'EPG source' — make it EPG-only
## Summary

The system-managed channel-source group ("Dispatcharr Channels", `is_channel_source=1`, epic 183.9) is created and re-synced with **both** `epg_match_enabled=1` and `team_streams_enabled=1`. Its display name is "Dispatcharr Channels (EPG source)" and the feature is described as matching each channel's own EPG, but Team matching runs alongside EPG.

## Impact

Single-team RSN streams curated onto Dispatcharr channels (e.g. "SPORTSNET NEW YORK HD") get fuzzy-matched as a team stream and fan out across that team's event channels — even though the source is intended to be EPG-driven. Observed in the wild: the channel-source group shows ~471 `fuzzy` (team) matches alongside its EPG matches.

Because `ensure_channel_source_group()` force-writes `team_streams_enabled=1` on **every** generation run, users cannot disable Team matching on this group via the per-group toggles — it self-heals back on. The only current workaround is priority rules to push the RSN down the source queue.

## Expected

The channel-source group should match **EPG only** (`epg_match_enabled=1`, `team_streams_enabled=0`, `name_match_enabled=0`), matching what its name and docs already claim.

## Location

`teamarr/database/groups.py::ensure_channel_source_group` (both the UPDATE re-sync branch and the create branch hardcode `team_streams_enabled=1`).

## Design note for reviewer

Straightforward fix is to drop `team_streams_enabled` to 0 for this group. Open question worth a reviewer decision: should Team matching on the channel source be fully removed, or made opt-in via a setting for anyone who relies on the team fan-out? Leaning EPG-only per the feature's stated intent.

lpukatch:
Closes #406 (lands on dev; issue closes at release).

## Problem

The system-managed channel-source group (`is_channel_source=1`, "Dispatcharr Channels", epic 183.9) was created **and re-synced every generation run** with both `epg_match_enabled=1` **and** `team_streams_enabled=1`. The feature is documented as EPG-driven ("each channel's own EPG"), and its display name is "(EPG source)" — but Team matching ran alongside EPG.

Effect: single-team RSN streams curated onto Dispatcharr channels (e.g. "SPORTSNET NEW YORK HD") got **fuzzy-matched as team streams** and fanned out across that team's event channels. In one live DB the channel-source group showed ~471 `fuzzy` (team) matches next to its EPG matches. And because `ensure_channel_source_group()` force-writes the flag on every run, users couldn't turn Team matching off via the per-group toggles — it self-healed back on. The only workaround was priority rules.

## Fix

Set `team_streams_enabled=0` on the channel-source group in **both** branches of `ensure_channel_source_group` (create + re-sync UPDATE). The source now matches by **EPG program data only** — matching the documented intent.

- Existing databases self-correct on the next generation run via the UPDATE branch; stale team (`fuzzy`) matches in that group are cleaned up by reconciliation.
- Regression assertions added to `test_ensure_channel_source_group_idempotent_and_synced` covering both the create and re-sync paths.
- Docs clarification in `program-matching.md` stating channel-source matches by EPG only, never by stream name / team inference.

## Reviewer decision requested

This makes the channel source **strictly** EPG-only. Alternative would be to keep Team matching available as an **opt-in setting** for anyone relying on the single-team fan-out. I went EPG-only per the feature's stated intent, but flagging in case you'd prefer the setting-gated approach. Leaving this at PR for your review rather than self-merging.

## Gates

- `pytest tests/` — 1673 passed
- No frontend changes (build unaffected)


egyptiangio (approved):
## Review — approved, going strict EPG-only

Decision on the flagged question: **strict EPG-only, as written.** The feature's name, docs, and the 183.9 epic intent all say EPG-driven; the team fan-out was emergent behavior, not design — and the force-resync meant users couldn't even opt out of it, which makes it a bug regardless. If someone was relying on RSN fan-out through this source, the right home for that is a regular team-matching source; we'll carry a changelog note. No opt-in setting until someone actually asks for it.

Verified beyond the diff:

- `skip_builtin_filter=1` + `epg_match_enabled=1` still short-circuit the builtin pre-filter in `stream_fetcher`, so linear/RSN streams keep reaching the EPG matcher — the `team_streams_enabled` league-hint exemption in `stream_filter` is moot for this group.
- The matcher-side gate (`TEAM_ONLY` → `team_streams_disabled`) is the only behavioral change; the EPG index/resolver path is untouched.
- Both branches (create + UPDATE re-sync) covered, and the re-sync test asserts the flags self-correct on existing DBs — exactly the self-healing property that made the bug sticky in the first place.

CI green. Merging — thanks for the clean, well-scoped fix and for flagging the design question instead of burying it.

PR #476, opened 2026-07-17
Racing EPG matching fixes from the Belgian GP weekend: programme dates, per-session plans, session scoping, archival rejection; loopback EPG resolution

tomwinterrose:
Ports a weekend's worth of live-verified racing/EPG matching fixes from my apex fork, all found and validated during the 2026 Belgian GP + IndyCar Nashville + NASCAR Trucks weekend (2026-07-17). Follow-up to #394/#405. Every fix below shipped on my production instance first; the "live" notes are actual observed failures, not hypotheticals.

## Commit 1 — Racing EPG matching (racing_matcher, racing_segments, matcher)

- **Programme-date matching**: EPG programmes carry no date in their text, so the racing matcher matched them against the *run's* target date — Friday's "Belgian Grand Prix: 1st Practice" listing stayed unmatched until the run date reached the weekend. The EPG path always has the programme's broadcast instant, so its own date is now the fallback. Name-path behavior unchanged (regression test included).
- **Per-session plan buckets**: `_compute_epg_plan` kept one programme per *event*, but a race weekend's guide has one programme per *session*, all matching the same event — FP1/FP2/Quali/Race collapsed to a single entry/window. Racing entries now bucket by (event, nearest session).
- **Programme-slot session scoping**: EPG-matched linear channels took the generic fan-out (which excludes practice), so FP1 aired with no channel despite the guide naming it. EPG entries now scope to the session(s) their programme slot covers — practice included, since the listing is positive evidence of coverage.
- **Session-named streams with no such session are dropped**: ESPN's IndyCar events carry only the race session; dedicated "Practice 1:"/"Qualifying:" feeds fell back to the full fan-out and landed on the Race channel. Also handles Apple-TV-PPV names ("NEXT | BELGIUM: RACE | …") where the session is a colon-value in a middle segment — those race feeds were landing on the Qualifying channel.
- **Archival-replay rejection**: Plex's FAST "F1 Channel" aired *2008 Belgian Grand Prix* 33 minutes into 2026 qualifying — inside anchor tolerance, with racing text evidence and a near-perfect fuzzy score. A year in the text that matches no candidate event's year now rejects the match (and invalidates stale cache entries). 1950–2049 only, so "2160P" never reads as a year.
- **Compound-spelling tolerance**: guide "Faith **Fest** 250" vs event "**FaithFest** 250 presented by Mercer Transportation" shared only "250" (46.6 < the 50 sanity threshold), so the Trucks qualifying broadcast never matched. Scoring now also compares bigram-augmented forms (adjacent tokens fused) and takes the max; unrelated pairs score *lower* augmented, so nothing previously rejected flips.
- **Strategy-1 text-evidence gate**: single-covering-event matches now also require `has_racing_text_evidence`, not just a score — generic venue words ("Speedway") could clear the sanity threshold on one shared token (live: a FIM Speedway GP stream binding to NASCAR's Atlanta event). This was already in apex but had never been upstreamed.

## Commit 2 — EPG resolver: loopback streams + generated-guide poisoning

- **Loopback-stream resolution**: a "Dispatcharr inside Dispatcharr" M3U (one stable stream per curated channel — one match entry per event instead of dozens of provider duplicates) was unreachable by all three resolver paths: the tvg_id is a channel *number*, names are ambiguous across guides, and channel-membership curation dies on every refresh because the loopback account recreates its streams with new ids. The stream's URL (`/proxy/ts/stream/<channel uuid>`) is its stable identity; new step 1b resolves that uuid to the source channel's EPG link. The uuid map rides on the existing paginated channel fetch — zero extra API calls.
- **Own-guide poisoning**: a stream that EPG-matches gets consolidated into the managed event channel it matched; next run the channel path resolves it to our own generated guide (whose programmes the index excludes) and the stream never matches again. Own source is identified by the `dispatcharr_epg_id` setting rather than the `"_Teamarr"` name (which misses renamed sources). Own-guide links now fall through the cascade.
- **Sibling-install poisoning**: an apex/vroomarr install writing to the same Dispatcharr has the same effect with worse symptoms — its synthetic guide airs all-day "Coming up: F1 Racing …" placeholder blocks, which bound TSN 5 (a qualifying+race broadcaster) to all five Belgian GP session channels. Synthetic event guides are recognized by their code-controlled tvg prefixes (`teamarr-event-`, `apex-event-`) regardless of source id.

## Tests

Full suite: **1990 passed** on this branch. ~35 new tests covering every fix above, each documenting the live failure it pins.

I'm aware of the ongoing discussion about motorsports scope — happy to split this PR (the commit-2 resolver fixes are general EPG matching, not racing-specific) if that makes review or any future extraction easier.

### DIFF
--- a/teamarr/consumers/event_group_processor/matching.py
+++ b/teamarr/consumers/event_group_processor/matching.py
@@ -191,7 +191,9 @@ class StreamMatching:
         # curated channel fallback). Both are single scoped fetches.
         try:
             epg_data_list = self._dispatcharr_client.channels.get_epg_data_list()
-            stream_channels = self._dispatcharr_client.channels.get_stream_channel_map()
+            stream_channels, channel_by_uuid = (
+                self._dispatcharr_client.channels.get_channel_maps()
+            )
         except Exception as e:
             logger.warning("[EPG-MATCH] Failed to load EPG resolution data: %s", e)
             return None
@@ -201,7 +203,10 @@ class StreamMatching:
         # excluded so we never resolve a stream to our generated guide.
         active_source_ids = self._active_epg_source_ids()
         resolution, _stats = resolve_program_tvg_ids(
-            streams, epg_data_list, stream_channels, active_source_ids=active_source_ids
+            streams, epg_data_list, stream_channels,
+            active_source_ids=active_source_ids,
+            channel_by_uuid=channel_by_uuid,
+            own_source_id=self._own_epg_source_id(),
         )
 
         # Window mirrors the event match window so programs overlapping any
@@ -241,6 +246,21 @@ class StreamMatching:
             group.id, index.program_count(), len(index.tvg_ids()),
         )
         return index
+
+    def _own_epg_source_id(self) -> "int | None":
+        """The app's OWN configured EPG-source id (``dispatcharr_epg_id`` setting).
+
+        Resolved at runtime rather than assumed: name checks ("_Teamarr")
+        silently miss installs whose source was renamed.
+        """
+        try:
+            from teamarr.database.channels.settings_helpers import get_dispatcharr_settings
+
+            with self._db_factory() as conn:
+                return get_dispatcharr_settings(conn).get("epg_id")
+        except Exception as e:
+            logger.debug("[EPG-MATCH] could not resolve own epg_id setting: %s", e)
+            return None
 
     def _active_epg_source_ids(self) -> set[int] | None:
         """Enabled EPG-source ids for name/direct matching (excludes _Teamarr).
--- a/teamarr/consumers/event_group_processor/stream_fetcher.py
+++ b/teamarr/consumers/event_group_processor/stream_fetcher.py
@@ -80,6 +80,9 @@ class StreamFetcher:
                     "name": s.name,
                     "tvg_id": s.tvg_id,
                     "tvg_name": s.tvg_name,
+                    # Loopback EPG resolution reads the source-channel uuid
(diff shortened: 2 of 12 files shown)


egyptiangio (approved):
Reviewed the full diff. This is exemplary contribution work — every change is pinned to an observed live failure, documented where it lives, and covered by both positive and negative tests. Approving.

**What stands out:**
- Non-regression discipline is excellent: bigram augmentation uses `max()` so no previously-rejected pair can flip; the year gate is bounded 1950–2049 with an explicit "2160p never reads as a year" test; loopback resolution is an additive step 1b that channel-curation still outranks; own-guide detection keys off the `epg_id` *setting* rather than the "_Teamarr" name (which we know misses renamed sources).
- The uuid map riding the existing paginated channel fetch (zero extra API calls) is the right shape, and `get_stream_channel_map()` delegating to `get_channel_maps()[0]` keeps existing callers untouched.
- The archival-replay cache invalidation (deleting stale pre-gate cache binds) closes the loop most fixes of this kind forget.
- `uv.lock` version sync: good catch — our release bumps `pyproject.toml` without re-locking; you fixed our process miss for us.

**Minor notes, none blocking:**
- `_own_epg_source_id` opens a DB connection per `_build_epg_index` call — negligible at once-per-group-per-run, just noting it.
- The `apex-event-` sibling prefix hardcode is fine by me given it's your own install family; if a third fork appears we can make it configurable.
- No need to split the PR — the commit boundaries are clean enough to review as-is.

Merging. Thanks as always — the "live notes" style of PR description makes review dramatically cheaper.

PR #528, opened 2026-07-30
feat: EPG-resolver region-prefix/network-alias widening + shared text_primitives module

ndevries84:
Widens `epg_resolver.normalize_channel_name()`'s region-prefix stripping and adds a network-abbreviation alias table, then extracts the quality-token and region-code concepts both `epg_resolver.py` and `classifier.py` had independently hardcoded into one shared module (`text_primitives.py`) so a token or region code added once applies everywhere.

## What's included

1. **Region-prefix widening:** `_REGION_PREFIX` now also recognizes a spaced dash (`" - "`) as a grouping delimiter alongside `:` and `|` (`"US - ESPN"` → `"espn"`), and the country-code allowlist gained Asia-Pacific/Middle East codes (jp, kr, cn, za, sa, qa, ae, il, pk, ph, th, vn, my, sg, id, hk, tw). Guarded so undelimited text (`"USA ESPN"`), out-of-allowlist codes (`"LA - Lakers Live"`), and unspaced hyphens (`"FR-ESPN"`) are still left untouched.
2. **Network aliases:** `NETWORK_ALIASES` maps common abbreviations (fs1, fs2, nbcsn, cbssn, sn) to their full EPGData-style names, applied as the final step of `normalize_channel_name()`, matched on the whole normalized string or its leading token only (so `"sn 360"` aliases but `"wisconsin sports"` and `"espn2"` don't).
3. **Shared `text_primitives` module:** `QUALITY_TOKENS`, `TEAM_SAFE_QUALITY_TOKENS`, `REGION_CODES`, and `strip_quality_tokens()` become the single source both `epg_resolver.py`'s quality-token/region regexes and `classifier.py`'s `_clean_team_name` start/end quality-token strip now derive from, instead of each keeping its own hardcoded copy. `classifier.py` picks up hevc/h265/h264 as a result, while deliberately continuing to exclude hq/lq (too collision-prone for team/venue names — documented in the module docstring).

## Motivation

Kept as one PR because `text_primitives.py` is a genuine shared dependency: the region-prefix regex in `epg_resolver.py` and the quality-token strip in `classifier.py` both get rebuilt from it in the same logical change. Splitting the region/alias widening from the text_primitives extraction would leave an intermediate state where `epg_resolver.py` has two competing hardcoded token lists for one release; keeping them together avoids that churn and matches how the work was actually done and tested (tests were written against the shared module before either consumer was touched).

## What changed

- `teamarr/consumers/matching/epg_resolver.py` — widened region-prefix delimiter/allowlist, `NETWORK_ALIASES`, then re-derived `_QUALITY_TOKENS`/`_REGION_PREFIX` from `text_primitives`
- `teamarr/consumers/matching/text_primitives.py` — new shared module
- `teamarr/consumers/matching/classifier.py` — `_clean_team_name` quality-token strip now sourced from `TEAM_SAFE_QUALITY_TOKENS`

## Behavior-change callouts (worth flagging for maintainer sign-off)

- **Region-prefix stripping now applies to substantially more codes and a new delimiter shape.** Any provider feed using `"XX - Channel Name"` for a code in the widened allowlist, or one of the newly added Asia-Pacific/Middle East codes, will normalize differently than before — this is the intended fix, but it's a behavior change on live catalogs, not just new coverage of previously-untouched strings.
- **`NETWORK_ALIASES` is new normalization behavior**, not a bug fix per se — `"FS1"` and `"Fox Sports 1"` now collapse to the same normalized name where they previously didn't. Worth explicit sign-off since it changes EPG-resolution matching outcomes for those five abbreviations on any existing installation.
- No dry-run posture applies (pure string normalization, no destructive action).

Tests: `tests/epg/test_epg_resolver.py` (extended), `tests/test_text_primitives.py` (new) — `pytest tests/ -q`: 2092 passed (2035 baseline + 57 new); `ruff check teamarr/ tests/`: clean.

### DIFF
--- a/teamarr/consumers/matching/classifier.py
+++ b/teamarr/consumers/matching/classifier.py
@@ -16,9 +16,19 @@ from re import Pattern
 from typing import cast
 
 from teamarr.consumers.matching.normalizer import NormalizedStream, normalize_stream
+from teamarr.consumers.matching.text_primitives import TEAM_SAFE_QUALITY_TOKENS
 from teamarr.services.detection_keywords import DetectionKeywordService
 
 logger = logging.getLogger(__name__)
+
+# Quality/format tokens safe to strip from an extracted TEAM name, at the
+# start or end of the string only. Sourced from the shared
+# text_primitives.TEAM_SAFE_QUALITY_TOKENS (a conservative subset of
+# QUALITY_TOKENS -- see that module for the hq/lq exclusion rationale), so a
+# token added there automatically applies here too.
+_TEAM_QUALITY_ALTERNATION = "|".join(re.escape(token) for token in sorted(TEAM_SAFE_QUALITY_TOKENS))
+_TEAM_QUALITY_START = re.compile(rf"^\s*\b(?:{_TEAM_QUALITY_ALTERNATION})\b\s*", re.IGNORECASE)
+_TEAM_QUALITY_END = re.compile(rf"\s+\b(?:{_TEAM_QUALITY_ALTERNATION})\b\s*$", re.IGNORECASE)
 
 
 class StreamCategory(Enum):
@@ -945,9 +955,9 @@ def _clean_team_name(name: str) -> str:
         flags=re.IGNORECASE,
     )
 
-    # Remove HD, SD, 4K, UHD quality indicators (at start or end)
-    name = re.sub(r"^\s*\b(HD|SD|FHD|4K|UHD)\b\s*", "", name, flags=re.IGNORECASE)
-    name = re.sub(r"\s+\b(HD|SD|FHD|4K|UHD)\b\s*$", "", name, flags=re.IGNORECASE)
+    # Remove quality indicators (at start or end) -- team-name-safe subset only
+    name = _TEAM_QUALITY_START.sub("", name)
+    name = _TEAM_QUALITY_END.sub("", name)
 
     # Remove broadcast network indicators like (CBS), (FOX), (ABC), (NBC), (ESPN)
     name = re.sub(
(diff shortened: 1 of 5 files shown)


egyptiangio:
Declining as-is — not on code quality, but on need: no part of this traces to an observed failure.

- **Region widening + `NETWORK_ALIASES`:** tier-3 name matching is deliberately strict because a wrong guess silently attaches wrong guide data. Widening it starts from a real case: which provider/catalog produced names the current normalization missed, and what did the failed resolution look like? File that as an issue — it becomes the justification and the test fixture for exactly as much widening as the case requires. (If the aliases return with evidence, one defect to address: a catalog carrying both `"FS1"` and `"Fox Sports 1"` as separate EPGData rows collapses to one normalized name → ambiguity-skip → a stream that strict-matches today silently loses guide data. Needs a test and a decided behavior.)
- **`text_primitives` extraction:** nothing has drifted and no bug traces to the duplicated lists — and it isn't behavior-neutral (`_clean_team_name` would gain `hevc`/`h265`/`h264` stripping). Restructuring working matcher internals needs a problem to justify it.

Process: PRs here base on `dev`, not `main` — please retarget (Edit → base branch; no re-push needed). This repo is issue-first: the problem statement lands as an issue before the implementation.

ndevries84:
Agreed on both points. Nothing here traces back to an actual failed resolution, and you're right the `text_primitives` split isn't behavior-neutral once the `hevc`/`h265`/`h264` stripping in `_clean_team_name` is accounted for.

Withdrawing the region/alias widening and the text_primitives extraction. If I run into a real feed name that current normalization mis-resolves, I'll file it as an issue with the actual string before touching the matcher again.

Closing this.

PR #529, opened 2026-07-30
fix: separator earliest-match, numeric event_id sort, generalized mojibake repair

ndevries84:
Three small, independent correctness fixes in the stream-matching pipeline, each landed behind its own failing-tests-first commit.

## What's fixed

1. **`DetectionKeywordService.find_separator` now picks the earliest-position separator match** instead of the first one found by walking `GAME_SEPARATORS` in priority order. `"Norwich at Home vs Ipswich"` was wrongly splitting on `" vs "` (later in the string, but higher priority) instead of `" at "` (earlier). List order and match length now only break same-position ties.
2. **`get_all_channels_sorted`'s tie-break sorted `event_id` as a string**, so `"10"` sorted before `"2"`. Added `_event_id_sort_key()`: numeric ids compare numerically and sort ahead of non-numeric ids; two non-numeric ids still fall back to string comparison.
3. **`fix_mojibake()` only repaired a hard-coded list of German/Spanish/French double-encoding patterns**, missing Nordic/Polish characters entirely, and `m3u.py`'s `_fix_double_encoded_utf8()` duplicated an *unguarded* latin-1/UTF-8 round trip with no protection against producing U+FFFD. Added `try_fix_double_encoded()`: a guarded, generic round-trip repair that rejects on exception or a U+FFFD in the result. `fix_mojibake()` tries it first, falling back to the old pattern list only as a no-op guard; `m3u.py` now delegates to the same function instead of re-implementing the recode.

## Motivation

All three are real-world stream-name defects found while working the matching pipeline: separator misdetection produces wrong team splits, lexicographic event_id sort produces a visibly wrong channel-number order once ids hit double digits, and narrow mojibake patterns leave Nordic/Polish provider feeds garbled. None change any public API shape — same inputs/outputs, corrected internal logic.

## What changed

- `teamarr/services/detection_keywords.py` — `find_separator` earliest-match selection
- `teamarr/database/channel_numbers.py` — `_event_id_sort_key()` helper, numeric-aware tie-break
- `teamarr/consumers/matching/normalizer.py` — `try_fix_double_encoded()`, `fix_mojibake()` now tries it first
- `teamarr/dispatcharr/managers/m3u.py` — `_fix_double_encoded_utf8()` delegates to the shared function

## Behavior-change callouts

- Separator selection can change which text a stream name is split on for names where a lower-priority separator occurs earlier than a higher-priority one — intentional, but worth a maintainer's eyes since it changes team-extraction output for that (narrow) case shape.
- No dry-run/preview posture applies here (these are pure-function matching fixes, not destructive operations).

Tests: `tests/test_separator_position_priority.py`, `tests/test_channel_event_id_sort.py`, `tests/test_mojibake_repair.py` — `pytest tests/ -q`: 2050 passed (2035 baseline + 15 new); `ruff check teamarr/ tests/`: clean.

### DIFF
--- a/teamarr/consumers/matching/normalizer.py
+++ b/teamarr/consumers/matching/normalizer.py
@@ -79,8 +79,54 @@ MOJIBAKE_PATTERNS = [
 ]
 
 
+def try_fix_double_encoded(text: str) -> str:
+    """Attempt to repair double-encoded UTF-8 text via a latin-1/utf-8 round trip.
+
+    Mojibake like "MÃ¼nchen" (for "München") typically happens when UTF-8
+    bytes get mis-decoded as latin-1 somewhere upstream (latin-1 is a
+    lossless 1:1 byte<->codepoint mapping, so no information is lost -- it
+    can be reversed). Reversing it means re-encoding the broken text back to
+    latin-1 bytes, then decoding those bytes as UTF-8.
+
+    This is intentionally conservative and safe to call on arbitrary,
+    possibly-already-correct text: if the round trip raises (the text wasn't
+    actually latin-1-of-utf8, e.g. legitimate "SÃO PAULO FC") or produces a
+    U+FFFD replacement character (a silently swallowed decode error), the
+    original text is returned unchanged rather than mangled. It is also
+    idempotent -- re-running it on already-correct text is a no-op, since
+    re-encoding proper unicode text as latin-1 generally fails or decoding
+    the result as UTF-8 does.
+
+    Args:
+        text: Potentially double-encoded text
+
+    Returns:
+        The repaired text, or the original text unchanged if the round trip
+        wasn't safely reversible.
+    """
+    if not text:
+        return text
+
+    try:
+        candidate = text.encode("latin-1").decode("utf-8")
+    except (UnicodeEncodeError, UnicodeDecodeError):
+        return text
+
+    if "�" in candidate:
+        return text
+
+    return candidate
+
+
 def fix_mojibake(text: str) -> str:
-    """Fix common mojibake patterns from double-encoded UTF-8.
+    """Fix mojibake (double-encoded UTF-8) in text.
+
+    Tries the generic, guarded latin-1/utf-8 round trip first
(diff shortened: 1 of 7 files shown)


egyptiangio:
One accepted, two held for evidence:

**✅ Guarded mojibake repair — accepted.** This one fixes a demonstrable defect: our current pattern table maps `Ã¯`/`Ã´`/`Ã¹` to bare letters instead of `ï`/`ô`/`ù`, and `m3u.py` carries an unguarded duplicate of the same round trip. Verified legit-diacritic names (`"SÃO PAULO FC"`) pass through untouched. Split it into its own PR based on `dev` and it merges — or I'll take-and-fix with credit in the changelog.

**⏸ Numeric `event_id` sort — no observed problem.** It's the fifth tie-break in the sort key: it only fires for same-league channels with identical start times, its only job is determinism, and ESPN's fixed-length ids make string and numeric order identical anyway. If a real install shows visibly wrong ordering from it, file that.

**⏸ Separator earliest-match — no observed problem, and it's a trade, not a strict fix.** `at`/`@` appear early in real stream names as time/date/venue markers (`"Football at 8PM: Eagles vs Cowboys"`, `"… @ Jun 20"`), and the current priority order is what keeps `vs` winning those; earliest-position-wins breaks that shape to fix another. Show actual stream names from a real catalog the current order mis-splits — as an issue — and the design happens against that data. (The same-position longer-match tie-break, `" vs. "` over `" vs "`, is correct and welcome to ride along with a future evidenced change.)

Same process notes: retarget to `dev`, issue-first.

ndevries84:
Split the mojibake fix out to #537 based on dev, as you suggested.

On the separator change, I traced a real example through `extract_teams_from_separator` and earliest-match splits "Norwich at Home vs Ipswich" on " at ", which corrupts the second team name. So the case argues against my own change. You're right that `at`/`@` show up as venue and time markers more often than as a real separator. Withdrawing that piece.

Also dropping the `event_id` sort since I don't have a real install where it produced visibly wrong ordering.

Closing this out. Will come back issue-first with actual stream names if either one resurfaces with real evidence behind it.

PR #530, opened 2026-07-30
feat: suspect-duplicate detection safety net for channel reconciliation (Phase 3b, item 12)

ndevries84:
Teamarr dedupes channels only via a DB unique index on `(event_id, provider, keyword, stream_id)`. When name/team matching resolves two differently formatted streams of the same real game to different (or missing) event_ids, no duplicate-channel detection exists anywhere. This PR adds `find_suspect_duplicates`, a pairwise scan over active channels that flags a pair when: sport matches, event_id doesn't already tie them together, event windows overlap (or are unavailable, in which case the check is bypassed rather than blocking), and normalized-name similarity (via `rapidfuzz.fuzz.token_set_ratio` on `fuzzy_match.normalize_text`) clears a threshold (default 85.0).

This is **detection only** — a pure function, no DB access, no auto-merge. It's wired into `ChannelReconciler` as a new Step 3b alongside the existing exact-event_id duplicate check, reported as info-severity, non-auto-fixable `ReconciliationIssue`s (`issue_type="suspect_duplicate"`), wrapped in try/except so a detection bug can never break the rest of reconciliation, and skipped outright for fewer than 2 active channels.

## Motivation

Closes a real blind spot in duplicate detection that the DB unique index structurally cannot catch (two differently-named/keyworded streams of one game will always get distinct DB rows even if they're the same broadcast). Deliberately scoped to detection/reporting only for this PR — auto-merge is explicitly deferred to a later item, called out in the module docstring.

## What changed

- `teamarr/consumers/enforcement/duplicate_detector.py` — new `find_suspect_duplicates()`
- `teamarr/consumers/enforcement/__init__.py` — export `find_suspect_duplicates`
- `teamarr/consumers/reconciliation.py` — new `ChannelReconciler._detect_suspect_duplicates`, wired in as Step 3b; `"suspect_duplicate"` added to the issue-count summary dict

## Behavior-change callouts

- Purely additive/detection-only — no existing behavior changes, no auto-fix, no dry-run posture needed (nothing is mutated). The only observable change is new `info`-severity log lines (`issue_type="suspect_duplicate"`) appearing in reconciliation runs where a suspect pair is found; flagging this so it's not mistaken for noise by anyone monitoring reconciliation log volume.
- The similarity threshold (85.0) is a judgment call, not derived from any real-world false-positive/negative measurement — happy to tune it based on maintainer feedback.

Tests: `tests/test_duplicate_detector.py` — full behavioral contract covering event_id equality exclusion, sport equality, half-open window overlap with documented bypass, similarity threshold, deterministic pair ordering. `pytest tests/ -q`: 2056 passed (2035 baseline + 21 new); `ruff check teamarr/ tests/`: clean.

### DIFF
--- a/teamarr/consumers/enforcement/__init__.py
+++ b/teamarr/consumers/enforcement/__init__.py
@@ -12,6 +12,10 @@ consistency across channels:
 3. KeywordOrderingEnforcer: Ensures main channel has lower number than
    keyword channels for the same event (main before Spanish/French)
 
+4. find_suspect_duplicates: Detection-only safety net for channels that
+   the DB unique index can't catch -- two differently-formatted streams of
+   the same real game that never got the same event_id.
+
 Usage:
     from teamarr.consumers.enforcement import (
         KeywordEnforcer,
@@ -31,6 +35,7 @@ Usage:
 """
 
 from .cross_group import CrossGroupEnforcer, CrossGroupResult
+from .duplicate_detector import find_suspect_duplicates
 from .keywords import KeywordEnforcementResult, KeywordEnforcer
 from .ordering import KeywordOrderingEnforcer, OrderingResult
 
@@ -41,4 +46,5 @@ __all__ = [
     "CrossGroupResult",
     "KeywordOrderingEnforcer",
     "OrderingResult",
+    "find_suspect_duplicates",
 ]
--- a/teamarr/consumers/enforcement/duplicate_detector.py
+++ b/teamarr/consumers/enforcement/duplicate_detector.py
@@ -1,0 +1,114 @@
+"""Duplicate-channel detection safety net (Phase 3b, item 12).
+
+Teamarr dedupes channels only via a DB unique index on ``(event_id,
+provider, keyword, stream_id)``. When name normalization or fuzzy matching
+fails to resolve two differently-formatted streams of the SAME real game to
+one ``event_id``, a duplicate channel is created with no detection anywhere.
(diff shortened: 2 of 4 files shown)


egyptiangio:
Declining for lack of observed need:

- Zero reports of duplicate channels in this tracker's history.
- Ran your exact detection logic read-only against a production install: 29 active channels → 0 suspect pairs.
- Over the same install's full 17,564-channel history it flags ~1,700 pairs, nearly all legitimate (doubleheaders, men's/women's same-day games) — a report at that false-positive rate gets ignored before a real duplicate ever appears.
- Wiring defect besides: `_detect_suspect_duplicates` passes `event_end=None` for every channel, which per the function's own contract bypasses the window gate entirely — the time filter never runs.

If you have a real install where matcher misresolution produced duplicate channels, file it as an issue with the channel/stream data — that reopens this with a concrete fixture to design against.

Same process notes: retarget to `dev`, issue-first.

ndevries84:
Your false-positive read matches what we're seeing on our own data. Our first live run flagged 317 suspect pairs, and it was overwhelmingly same-day doubleheaders and same-day WTA matches, so the noise floor is exactly what you described.

Good catch on `event_end` always being `None` too, that means the time window was never actually gating anything. Pulled this out of our own build.

If we ever capture a real duplicate-channel case, I'll open an issue first and scope a version with an actual time window, which likely means storing `event_end` on managed channels. Closing this.

PR #537, opened 2026-08-01
fix: guarded mojibake repair

ndevries84:
Guarded, generalized double-encoded-UTF-8 (mojibake) repair, split out of #529 per maintainer request.

## The defect

`fix_mojibake()` only repaired a hard-coded list of German/Spanish/French double-encoding patterns — missing Nordic/Polish characters entirely — and some of those hard-coded mappings were themselves wrong (`Ã¯`/`Ã´`/`Ã¹` mapped to bare letters instead of `ï`/`ô`/`ù`). Separately, `m3u.py`'s `_fix_double_encoded_utf8()` duplicated an *unguarded* latin-1/UTF-8 round trip with no protection against producing U+FFFD replacement characters on bad input.

## The fix

- Added `try_fix_double_encoded()` in `normalizer.py`: a guarded, generic latin-1-encode/UTF-8-decode round trip that rejects on exception or a U+FFFD in the result, leaving the input untouched otherwise.
- `fix_mojibake()` now tries the generic repair first, falling back to `MOJIBAKE_PATTERNS` only as a no-op guard.
- `m3u.py`'s `_fix_double_encoded_utf8()` now delegates to the same shared function instead of re-implementing the recode (using a deferred/local import to avoid a circular import back through `teamarr.dispatcharr.factory`).

Verified legitimate diacritic names (e.g. `"SÃO PAULO FC"`) pass through untouched, per the maintainer's own check.

## What changed

- `teamarr/consumers/matching/normalizer.py` — `try_fix_double_encoded()`, `fix_mojibake()` now tries it first
- `teamarr/dispatcharr/managers/m3u.py` — `_fix_double_encoded_utf8()` delegates to the shared function

## Tests

- `tests/test_mojibake_repair.py` (new) — round-trip repair contract: never returns U+FFFD, leaves legitimate non-mojibake text alone, is idempotent, and covers Nordic/Polish characters alongside the original German/Spanish/French cases.

`pytest tests/ -q`: 2168 passed. `ruff check teamarr/ tests/`: clean.

## Context

This is the mojibake portion of #529, split into its own PR based on `dev` per review feedback: "This one fixes a demonstrable defect... Split it into its own PR based on dev and it merges." The other two items from #529 (separator earliest-match, numeric event_id sort) are intentionally excluded — they're held pending real-world evidence per the same review and are not part of this PR.

### DIFF
--- a/teamarr/consumers/matching/normalizer.py
+++ b/teamarr/consumers/matching/normalizer.py
@@ -79,8 +79,54 @@ MOJIBAKE_PATTERNS = [
 ]
 
 
+def try_fix_double_encoded(text: str) -> str:
+    """Attempt to repair double-encoded UTF-8 text via a latin-1/utf-8 round trip.
+
+    Mojibake like "MÃ¼nchen" (for "München") typically happens when UTF-8
+    bytes get mis-decoded as latin-1 somewhere upstream (latin-1 is a
+    lossless 1:1 byte<->codepoint mapping, so no information is lost -- it
+    can be reversed). Reversing it means re-encoding the broken text back to
+    latin-1 bytes, then decoding those bytes as UTF-8.
+
+    This is intentionally conservative and safe to call on arbitrary,
+    possibly-already-correct text: if the round trip raises (the text wasn't
+    actually latin-1-of-utf8, e.g. legitimate "SÃO PAULO FC") or produces a
+    U+FFFD replacement character (a silently swallowed decode error), the
+    original text is returned unchanged rather than mangled. It is also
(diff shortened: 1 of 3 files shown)


egyptiangio:
Reviewed: the guarded round trip is correct (U+FFFD rejection + exception guard), the `SÃO PAULO FC` pass-through case holds, and the wrong hard-coded mappings (`Ã¯`→`i`, `Ã´`→`o`, `Ã¹`/`Ã»`→`u`) are now superseded by the generic repair. Verified locally on top of dev: ruff clean, 2168 tests pass. CI green. Merging — thanks @ndevries84.

Issue #569, opened 2026-08-15
[Bug]: Same-nickname teams crosstalk across leagues (SF Giants streams land on NY Giants channels)


lpukatch:
## Problem

`US: San Francisco Giants` (a TEAM_ONLY branded stream) is being attached to NFL `MIN/NYG` channels alongside the correct MLB `COL/SF` channels.

## Root cause

`_best_name_score()` in `teamarr/consumers/matching/team_matcher.py` scores a stream candidate against `max(token_set_ratio(stream, team.name), token_set_ratio(stream, team.short_name))`. The `short_name` leg was added in #480 so ESPN nicknames that share no words with the full name ("D-backs" vs "Arizona Diamondbacks") still match.

But ESPN gives both the NFL and MLB Giants `short_name = "Giants"`, and `token_set_ratio` returns **100** whenever the short name is a token subset of the stream candidate:

```
token_set_ratio("san francisco giants", "new york giants") =  57   # correctly rejects
token_set_ratio("san francisco giants", "giants")          = 100   # short_name leg wins
```

So the discriminating location is thrown away and any two teams sharing a nickname become interchangeable.

## Scope

This is systemic, not Giants-specific. Sweeping a real 10,320-row `team_cache`, the short_name leg lifts **2,015** cross-team comparisons (788 unique name pairs) from below-threshold to >= 85. Other examples: Racing Santander vs Racing Club, Estudiantes de La Plata vs Estudiantes de Merida, and every FIFA age-group pair (Italy U20 vs Italy U17, Brazil U17 vs Brazil U23, ...).

## Proposed fix

Gate the `short_name` leg: when it would beat the full-name score, require the stream's residual tokens (stream tokens minus short-name tokens) to be consistent with the team's own discriminator (name tokens minus short-name tokens) — allowing the discriminator's own tokens, its initialism, and the team abbreviation.

Measured on the same 10,320-team cache: 2,005 of 2,015 false positives eliminated, 9 regressions out of 29,900 self-match variants (all synthetic multi-token college abbreviations like "L-MD").

```
'san francisco giants' vs New York Giants:      before=100  after=57
'san francisco giants' vs San Francisco Giants: before=100  after=100
'sf giants'  vs New York Giants:                before=100  after=80
'sf giants'  vs San Francisco Giants:           before=100  after=100
'ny giants'  vs New York Giants:                before=100  after=100
'ny giants'  vs San Francisco Giants:           before=100  after=80
```

[...]

Issue #480, opened 2026-07-17
Nickname short names (D-backs) never fuzzy-match; DATE_MISMATCH reason over-claims


egyptiangio:
From a remote user's debug log (v2.12.x): `MLB 13 | Cardinals x D-backs start:2026-07-18 02:40` classifies cleanly (teams=Cardinals/D-backs, date trusted via ISO custom regex) but fails with `reason=date_mismatch` while other streams in the group match.

Two defects:

1. **Fuzzy scoring ignores `short_name`.** "D-backs" normalizes to tokens `{d, backs}` — zero overlap with "Arizona Diamondbacks", scores ~50 (< 60 threshold). ESPN's own `short_name` for the team IS "D-backs"; the scorer never looks at it. Fix: score against the best of name/short_name.

2. **`DATE_MISMATCH` over-claims (#474 fallout).** `date_rejected` increments for every candidate outside ±1 day — for an MLB stream with a trusted date that's nearly the whole window, so any failure gets labeled `date_mismatch` even when the real cause is the teams (as here — the date is actually fine: 02:40 UTC = 7:40 PM Arizona the previous evening, absorbed by the ±1 tolerance). Fix: score teams first, gate on date second — only count date-rejections for candidates whose teams would have matched.

egyptiangio:
Landed on dev via PR #481 (28104526). The reporting user's 'D-backs' alias will now fire, the stream also matches via short_name without any alias, and match failures log a [NEAR_MISS] diagnostic line. Closes at next release.

egyptiangio:
Round 2 landed on dev in 82ef9735 (PR #483): alias resolution is now PER SIDE — the reporting user's single 'd-backs' alias is sufficient ('Cardinals' matches by fuzz). Root cause of round 1's miss: the alias path required both teams aliased, and my test had fabricated the short_name. Their stream should match on the next run after updating.

egyptiangio:
One more on dev in fe2f290f (PR #490): **`d-backs` and `dbacks` are now built-in aliases** — the stream will match out of the box on the next dev image, no personal alias needed. (Also fixed alias precedence so a user's own alias for the same text still wins over the built-in.) For the reporting user: pull the next dev image and run — feed 67334 should attach to the event channel with no configuration.

egyptiangio:
Shipped in [v2.12.2](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.12.2).

PR #570, opened 2026-08-15
fix: gate short_name scoring so shared nicknames stop crossing leagues (#569)

lpukatch:
ESPN gives both the NFL and MLB Giants short_name 'Giants'. token_set_ratio
returns 100 whenever the short name is a token subset of the stream, so
'US: San Francisco Giants' scored 100 against the New York Giants (57 on the
full name) and the SF team-branded stream was attached to NFL channels.

The short_name leg (#480) now only stands in for the full-name score when the
stream's residual tokens agree with the team's own discriminator — its words,
its initialism, or the provider abbreviation. Bare-nickname streams are
unchanged; 'SF Giants' and 'NY Giants', ambiguous at 100 against both teams
today, now pin to the correct one.

Swept against a 10,320-row team_cache: 2,005 of 2,015 cross-team false
positives eliminated (788 unique name pairs, incl. Racing Santander/Racing
Club and every FIFA age-group pair), 9 regressions in 29,900 self-match
variants — all synthetic multi-token college abbreviations.

---

Bead: `teamarrv2-t7hc` · Issue: #569

### DIFF
--- a/teamarr/consumers/matching/team_matcher.py
+++ b/teamarr/consumers/matching/team_matcher.py
@@ -123,18 +123,69 @@ def _abbrev_equals(stream_code: str, event_abbrev: str | None) -> bool:
     return code == normalize_text(event_abbrev)
 
 
+def _initialism(tokens: list[str]) -> str:
+    """First letter of each token, in order ("san francisco" -> "sf")."""
+    return "".join(t[0] for t in tokens if t)
+
+
+def _short_name_leg_is_safe(stream_norm: str, name_norm: str, short_norm: str, abbrev: str) -> bool:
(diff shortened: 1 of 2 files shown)


egyptiangio:
Reviewed and verified locally: 11 new tests pass, full suite 2,211 green, ruff clean. The gate is exactly right — residual-token agreement with the team's own discriminator keeps #480's bare-nickname matching intact while restoring the location signal, and the sweep numbers (2,005/2,015 eliminated, 9 synthetic regressions) make the trade-off clear. The "SF Giants"/"NY Giants" disambiguation is a nice bonus on top of the fix. Merging — thanks @lpukatch.

PR #572, opened 2026-08-15
fix(tsdb): preserve stale schedules after rate limiting

### LINKED ISSUE #571: Preserve stale TSDB guide data after rate limiting
## Problem

Free-tier TheSportsDB schedule data can disappear from generated guide data when a cached response expires and TSDB continues responding with HTTP 429 after retries.

## Expected Behavior

When TSDB has a prior schedule response for the same request and fresh retrieval fails specifically because of persistent rate limiting, Teamarr should temporarily use that stale response until a fresh response succeeds.

## Scope

- Only retry-exhausted HTTP 429 responses trigger stale fallback.
- Ordinary transport or HTTP failures and legitimate empty schedules retain their current behavior.
- Fresh successful responses replace stale data immediately.

## Related Issues

- #217
- #181

sethwv:
## Summary

Free-tier TSDB leagues can lose guide entries when an expired schedule cache coincides with persistent HTTP 429 responses. Preserve the last successful raw schedule response for the same request and use it only after retries are exhausted.

Fresh TSDB responses immediately replace the stale fallback. Other HTTP failures, transport failures, and legitimate empty responses continue to return no data as before.

## Changes

- Add explicit expired-entry reads to `TTLCache`, isolated from normal cache hits.
- Surface retry-exhausted TSDB 429s as a distinct internal error.
- Use stale responses for `eventsday`, `eventsnextleague`, and `eventsseason` schedule fetches only after persistent 429s.
- Preserve existing behavior for TSDB lookup, validation, team, and non-rate-limit failure paths.
- Document temporary stale guide-data behavior for TSDB free-tier users.

Closes #571


sethwv:
Needs some time to test properly, let's see!

Issue #41, opened 2025-12-31
[Feature]: GUI Backup & Restore


jonzey231:
### Teamarr Version

V2 (alpha)

### Problem / Use case

Similar to how Dispatcharr now has a backup and restore option nested under Settings, that would be amazing here in Teamarr as well.

### Proposed solution

Create backup.
Download backup.
Upload backup.
Restore backup.

egyptiangio:
Implemented in 5cebfc5 - GUI backup & restore in Settings > Advanced tab

PR #587, opened 2026-08-22
feat: add opt-in generated sports previews

### LINKED ISSUE #613: Add opt-in generated sports previews from public ESPN variables
## Summary

Add an optional `{generated_preview}` template variable that produces a coherent, multi-sentence pregame description from structured, provider-reported facts.

This would complement—not replace—the existing `{game_preview}` variable. Provider editorial copy would remain the default behavior, and no starter template would enable generated prose automatically.

The initial implementation would provide sport-specific formatting for baseball, football, and basketball using ESPN data. Hockey and other sports would receive a generic matchup sentence. Dedicated support for those sports could be considered separately once their data contracts and presentation requirements are understood.

## Problem

ESPN’s `{game_preview}` value is often sparse or limited to a single line, such as:

> Boston Red Sox (72-59, third in the AL East) vs. Miami Marlins (67-65, third in the NL East)

Teamarr can retrieve additional structured facts, including:

- Team records
- Recent form
- Series status
- Probable starters
- Player statistical leaders
- Football team yardage
- Basketball points scored and allowed

Advanced users can compose descriptions from individual variables, but producing a robust multi-sport paragraph as a static template is difficult. Provider data is inconsistent and individual fields may be absent, leading to incomplete phrases, duplicate punctuation, or awkward sentence structure.

## Proposed behavior

Introduce:

- `{generated_preview}` — optional deterministic prose formatter
- `has_generated_preview` — condition indicating that enough data exists to generate a description

Neither would be added to starter/default template rows. Users would explicitly opt in by adding a condition such as:

- Condition: `has_generated_preview`
- Template: `{generated_preview}`

Existing installations and starter templates would continue using `{game_preview}` and Teamarr’s existing structured-preview fallback.

For baseball, football, and basketball, `has_generated_preview` would require meaningful structured pregame data. For other sports, it would require a complete matchup containing both team names.

## Public variable contract

All retrieved facts should be exposed as individually named variables. The generated formatter must compose the same typed fields rather than depend on an inaccessible data blob.

### Existing variables used

- `{home_team}`
- `{away_team}`
- `{venue}`
- `{home_team_record}`
- `{away_team_record}`
- `{home_last_five}`
- `{away_last_five}`
- `{series_summary}`
- `{home_team_ppg}`
- `{away_team_ppg}`

### New general variable

- `{week}`

### Baseball variables

- `{home_probable_starter}`
- `{away_probable_starter}`
- `{home_home_runs_leader}`
- `{away_home_runs_leader}`
- `{home_batting_average_leader}`
- `{away_batting_average_leader}`
- `{home_rbi_leader}`
- `{away_rbi_leader}`

### Football variables

- `{home_passing_leader}`
- `{away_passing_leader}`
- `{home_rushing_leader}`
- `{away_rushing_leader}`
- `{home_receiving_leader}`
- `{away_receiving_leader}`
- `{home_total_yards_per_game}`
- `{away_total_yards_per_game}`
- `{home_rushing_yards_per_game}`
- `{away_rushing_yards_per_game}`

### Basketball variables

- `{home_points_leader}`
- `{away_points_leader}`
- `{home_rebounds_leader}`
- `{away_rebounds_leader}`
- `{home_assists_leader}`
- `{away_assists_leader}`
- `{home_points_allowed_per_game}`
- `{away_points_allowed_per_game}`

Ambiguous convenience variables such as `{home_leader}` or `{away_leader}` should not be introduced. Each statistic should have an exact, documented meaning.

## Generated prose behavior

The formatter should:

- Use complete team names rather than provider abbreviations.
- Include only facts actually reported by the provider.
- Omit missing facts without producing broken punctuation or incomplete sentences.
- Never include betting information.
- Group multiple statistics for the same athlete cleanly.
- Produce deterministic output suitable for an EPG description.
- Preserve the retrieved pregame facts after kickoff so live box-score data cannot rewrite the channel description.
- Drop the preserved snapshot naturally when the channel advances to a different event.

### Baseball

Include, when available:

- Venue
- Series status
- Team records
- Recent form
- Home-run leader
- Probable starter with record and ERA

If a home-run leader is unavailable, use the batting-average leader and then the RBI leader as fallbacks.

Example:

> The Boston Red Sox visit the Miami Marlins at loanDepot park, with the Boston Red Sox leading the series 1-0. Boston enters at 72-59 after going 4-1 in its last 5 games, led by Willson Contreras with 26 home runs. Probable starter Payton Tolle is 8-6 with a 3.08 ERA. Miami enters at 67-65 after going 3-2 in its last 5 games, led by Heriberto Hernandez with 19 home runs. Probable starter Tyler Phillips is 3-6 with a 3.67 ERA.

### Football

Include, when available:

- Week and preseason designation
- Venue
- Team records
- Recent form
- Total yards per game
- Rushing yards per game
- Passing leader
- Rushing leader
- Receiving leader

All available leader categories should be included rather than treated as mutually exclusive fallbacks.

### Basketball

Include, when available:

- Venue
- Team records
- Recent form
- Points per game
- Points allowed per game
- Points leader
- Rebounds leader
- Assists leader

All available leader categories should be included. If the same athlete leads multiple categories, their statistics should be grouped rather than repeating the athlete’s name unnecessarily.

### Hockey and other sports

Sports without a dedicated formatter should receive a concise generic matchup sentence composed from the public team and venue variables.

Example:

> The New York Rangers visit the Boston Bruins at TD Garden.

The generic fallback should:

- Require both team names.
- Include the venue when available.
- Exclude records and statistics that do not have a supported sport-specific presentation contract.
- Never include betting information.
- Avoid applying baseball, football, or basketball wording to another sport.

This generic fallback would not constitute dedicated support for that sport. Sport-specific NHL or other-sport descriptions could be proposed separately later.

## Data lifecycle

Generated-preview fields represent a pregame snapshot.

Once an event begins:

- Existing pregame values must not be overwritten by live leaders or in-game team statistics.
- A refresh may update game status and scores without modifying the preserved preview facts.
- The snapshot should remain available for that event until the channel advances to its next event.
- Event and preview caches should serialize every typed field so values survive cache hits and restarts.

## Compatibility and scope

This feature would be entirely opt-in:

- `{generated_preview}` would not appear in starter/default rows.
- `has_generated_preview` would remain available in the condition selector.
- Existing ESPN-first behavior would remain unchanged.
- Advanced users could ignore the formatter and compose descriptions directly from the named variables.
- Provider-specific fields would resolve empty when the provider does not report them.
- No betting, odds, PickCenter, or predictor data would be used by the formatter.

The initial sport-specific contract would cover baseball, football, and basketball. Hockey and other sports would receive only the generic matchup fallback. Dedicated NHL support and other sport-specific expansions should be handled as separately justified enhancements.

## Expected support contract

The formatter should maintain:

- Core functionality for supported sports
- A complete generic matchup fallback for other sports
- Graceful handling of missing data
- Factual correctness as reported by ESPN
- Stable pregame snapshots after kickoff
- Exclusion of betting information

The formatter would not promise customizable tone, phrasing, sentence ordering, or unlimited sport-specific expansion. Users wanting complete control over wording and presentation should compose their own templates from the public named variables.

Minor wording refinements may be necessary after release as additional real-world payload variations are encountered, without changing the underlying public variable contract.

Because there are no NBA or NFL games available today for current end-to-end validation, the basketball and football formatters have been tested against representative ESPN payload structures and automated fixtures but will still require verification against live NBA and NFL pregame data as those seasons begin.

ncsufan8628:
Closes #613.

## Summary

- Parse ESPN pregame facts into typed `Event` fields and expose each fact as a named template variable.
- Add the opt-in `{generated_preview}` formatter and `has_generated_preview` condition.
- Compose generated prose exclusively from the same public typed fields available to user templates.
- Preserve complete pregame snapshots after kickoff so live box-score facts cannot rewrite EPG descriptions.

## Variable and formatter contract

The new public fields cover:

- Baseball: probable starters; home-run, batting-average, and RBI leaders.
- Football: week; passing, rushing, and receiving leaders; total and rushing yards per game.
- Basketball: points, rebounds, and assists leaders; points scored and allowed per game.

The implementation reuses existing public variables for teams, venue, records, recent form, series status, and basketball points per game. Missing provider facts render as an empty string. There is no free-form data blob and no ambiguous `{home_leader}` / `{away_leader}` alias.

Baseball, football, and basketball receive deliberately constrained sport-specific prose. Other sports receive a generic, complete matchup sentence when both teams are known, for example:

> The New York Rangers visit the Boston Bruins at TD Garden.

Hockey-specific prose is intentionally not inferred without representative in-season payloads. I would like to retain the option to propose NHL support separately once the season starts and real ESPN data can be validated. Until then, NHL remains covered by the generic fallback.

## Compatibility and safety

- `{generated_preview}` and `has_generated_preview` are registered but absent from every starter/default row.
- The existing ESPN-first priority ladder remains untouched; users must opt in through their own condition row.
- Existing `{game_preview}` and structured-preview behavior is unchanged.
- No fork-only starter migration, container workflow, or unrelated TSDB change is included.
- Odds, PickCenter, predictor, spreads, totals, and other betting data are excluded from parsing and rendering.
- Full provider team names are used instead of abbreviations in generated prose.

## Data lifecycle

The preservation work is isolated in commit `eb532355` for focused review. It expands the existing preview-field cache contract, retries sparse snapshots, serializes the typed fields, and keeps a completed pregame snapshot after kickoff. Event-ID scoping naturally drops that snapshot when a channel advances to its next game.

## Maintenance

I will own generated-prose wording issues and fixes under the contract in #613. Minor wording refinements may be necessary as additional real-world payload variations appear, without changing the public variable contract. Additional sport-specific coverage, including NHL, would come through a separately justified and validated enhancement.

## Validation

- Ruff: passed
- Pyright: 0 errors (one existing missing-source warning for `regex`)
- Backend: 2,386 tests passed
- Focused preview/default/registry/cache suite: 127 tests passed
- Coverage includes typed variable output, sparse-data degradation, default-template opt-in protection, betting exclusion, full team names, cache serialization, and freeze-after-kickoff behavior.


egyptiangio:
Thanks for the PR. Before review:

**1. Retarget to `dev`.** `main` only receives batched releases from `dev`. Change the base via the PR's Edit button.

**2. Open an issue first.** Every change here starts from a GitHub issue describing the problem — what's missing or broken today, for whom, and how the change addresses it — plus, for template variables, the contract: which facts, per sport. Link it from the PR.

**3. Most of this already exists (v2.10.0, #329),** on both `main` and `dev`:

- `{game_preview}` + `has_preview` — ESPN editorial preview, same-day
- `{home_last_five}` / `{away_last_five}` / `{last_five_summary}` + `has_structured_preview` — recent form from summary `lastFiveGames`, days-ahead
- `{series_summary}`, `{game_recap}`, `{game_event_note}`
- Starter templates already carry the tier ladder: `has_preview → {game_preview}`, then a `has_structured_preview` constructed-prose fallback
- `sports_data.py` `_PREVIEW_FIELDS` already preserves cached preview fields across refreshes

Refs: `teamarr/templates/variables/summary.py`, `teamarr/providers/espn/provider.py` (`_parse_last_five`), tvnk.14/15 notes at the top of `teamarr/database/default_templates.py`.

What "rich preview" adds is the same structured-preview idea with more facts: records, stat leaders, probable starters, team stats, week. That data is useful and worth exposing. The composed paragraph, the new variable/condition, and the starter-template rows are not — that's the existing feature under a new name.

**Scope that fits:**

- Parse the new facts into named fields on `Event` (beside `_parse_last_five`) and register each as a template variable — `{home_record}`, `{home_probable_starter}`, `{home_leader}`, etc. The `rich_preview_data: dict` blob doesn't work for this: the facts are locked inside it and only reachable through your renderer, so users can't use them individually.
- Drop `{game_preview_rich}`, `has_rich_preview`, `rich_preview.py`, and the starter-template changes. Composing those variables into prose is what a user's own template is for; Teamarr provides the variables, not the paragraph. (The starter rows also put generated prose at priority 5 above ESPN's editorial preview at 10, which inverts the ESPN-copy-first rule from tvnk.14.)

That's a much smaller diff and lands as variables every template can use.

ncsufan8628:
Thanks—that distinction makes sense. I agree that the newly retrieved facts should be exposed as individual named variables rather than being accessible only through an opaque `rich_preview_data` renderer.

I’d like to propose retaining `{game_preview_rich}` as an optional convenience variable, with these changes:

1. Retain `{game_preview_rich}` in the variable registry and `has_rich_preview` in the condition selector, but remove both from all starter/default template configurations. This would preserve Teamarr’s existing ESPN-first behavior. A user would need to explicitly opt in by adding a condition row or referencing the variable in a custom template.
2. Expose the underlying facts as named variables with a documented contract—for example, `{home_record}`, `{away_record}`, `{home_probable_starter}`, `{away_probable_starter}`, `{home_leader}`, `{away_leader}`, `{week}`, and applicable team-stat variables.
3. Refactor `{game_preview_rich}` to compose those same public fields rather than depending on an otherwise inaccessible data blob. Advanced users would have complete control through the individual variables, while users who want an automatic description could opt into the convenience renderer.

The reason for retaining the convenience variable is that a robust multi-sport paragraph is difficult to construct as a static template. Provider data is sparse and inconsistent: a probable starter, leader, record, recent form, venue, or particular statistic may be absent for any individual event. A static template can consequently produce dangling punctuation or incomplete phrases. The renderer conditionally includes only available facts, applies sport-appropriate wording, and produces a coherent description.

This would make `{game_preview_rich}` an optional formatting helper rather than a new default preview tier:

- Existing installations and starter templates continue using `{game_preview}` and the current structured-preview fallback.
- Nothing opts into rich generated prose automatically.
- Users wanting convenience can explicitly add `has_rich_preview → {game_preview_rich}` in the Conditions UI.
- Advanced users can use every underlying variable individually.

If this direction is acceptable, I’ll retarget the PR to `dev` and open an issue documenting the missing data, per-sport variable contract, availability rules, and intended fallback behavior before revising the implementation.

egyptiangio:
Agreed on the overall direction. Here are the terms that make it land:

**1. The facts become typed fields + named variables; the data blob goes away entirely.**

`rich_preview_data: dict` doesn't fit this codebase: every other fact on `Event` is a typed field, and every template variable reads one. A free-form dict whose schema is a private contract between `preview.py` and `rich_preview.py` is fragile — the renderer does no validation, so any shape drift prints raw dict text straight into EPG prose (verified locally by feeding it a slightly different `week` shape). Parse the facts into typed fields beside `_parse_last_five` and register each as a variable.

**Dedupe against the existing registry first.** Some of what the PR gathers is already exposed — `{home_team_record}`/`{away_team_record}` exist today and drive the current structured-preview starter rows. The genuinely new facts are leaders, probable starters, week, and per-team stats: `{home_leader}`, `{away_leader}`, `{home_probable_starter}`, `{away_probable_starter}`, `{week}`, and the stat variables. The issue should carry the contract: which facts, per sport, which existing variables they reuse, and what each new variable renders when data is absent (empty string, like existing variables).

**2. The convenience renderer can stay — renamed, as a formatter over those public fields, and contributor-maintained.**

Your sparse-data argument is valid for this engine: conditions select template rows, so users can't write phrase-level conditionals, and a hand-built template will produce dangling fragments on sparse days. A renderer that degrades gracefully earns its place under these constraints:

- **Rename `{game_preview_rich}` → `{generated_preview}` and `has_rich_preview` → `has_generated_preview`.** The registry already carries `{game_preview}` (ESPN's human-written copy), `has_preview`, and `has_structured_preview`; "rich" reads as a better `game_preview` when it's actually the machine-composed one. Provenance belongs in the name — it also self-documents the intended pecking order next to ESPN's copy.
- It composes **exclusively from the public typed fields** — one source of truth, the blob is deleted, and the renderer is testable against the same contract users see.
- Opt-in only: registered variable + condition, zero starter-template rows, priority ladder untouched. The intended slot is between `has_preview` (ESPN's words, the moment they publish) and `has_structured_preview` — the days-before-game window where ESPN hasn't written anything yet — but users place that row themselves.
- Wording stays scoped to the sports the ESPN summary actually serves (baseball, football, basketball). Anything else renders a generic-but-complete sentence — no speculative per-sport prose.

Being direct about the constraint that decides this: **I don't have the bandwidth to maintain generated prose.** The variables are ordinary surface — typed fields with tests — and I'll own those like any other part of the registry. The renderer is different: sentence-quality issues, per-sport wording requests, "this reads oddly for X" reports. So the arrangement for `{generated_preview}` is:

- **You own it.** Issues about its output get routed to you; wording fixes and any future sport coverage come from you.
- **Scope is frozen at merge.** The three sports above, boring sentence shapes, no dialect growth. Expansion PRs need the same case this one made.
- **If it becomes unowned, it gets frozen or removed, not maintained.** Because it's opt-in, absent from starters, and built on the public variables, removing it later breaks nothing except templates that explicitly opted in — users would fall back to composing from the same variables.

If that arrangement doesn't work for you, the variables-only scope from my first comment stands on its own and I'd merge it happily — that's where ~90% of the value is anyway.

**3. Mechanics before review:**

- Retarget to `dev`. Note that dev just landed #592, which rewrote the `sports_data.get_events` path your caching changes touch — rebase onto current dev first; the conflict is real.
- Issue first, linked from the PR (the variable contract above is most of its body, and it becomes the docs page).
- The freeze-after-kickoff and `_PREVIEW_FIELDS` changes are where review will concentrate — keep that work in its own commit if you can.

The odds-exclusion and freeze tests are the right instincts — keep those.

ncsufan8628:
Updated per your review terms:

- retargeted to `dev` and linked issue #613
- replaced the data blob with typed `Event` fields and exact named variables
- renamed the opt-in helper to `{generated_preview}` / `has_generated_preview`
- left every starter/default row and the ESPN-first priority ladder unchanged
- removed fork-only CI, private starter migration, and unrelated TSDB history
- limited sport-specific prose to baseball, football, and basketball; other sports use a complete generic matchup sentence
- isolated freeze-after-kickoff behavior in its own commit (`eb532355`)
- retained betting-exclusion and snapshot tests

I will own generated-prose wording maintenance under the contract in #613. NHL-specific prose is intentionally deferred until in-season ESPN payloads are available to validate; hockey uses the generic fallback meanwhile.

Fork CI passed Ruff/backend tests, Pyright, ESLint, and frontend build: https://github.com/ncsufan8628/teamarr/actions/runs/32909129462

ncsufan8628:
Added a small follow-up for NFL-specific wording quirks observed in a real ESPN preseason payload.

ESPN sometimes returns complete abbreviated stat lines inside fields named like yardage-only categories. That previously produced awkward generated prose such as `110 YDS passing yards` and `including 85 rushing`.

The formatter now expands those abbreviations only in `{generated_preview}` while leaving the public named variables’ provider facts intact. Examples:

- `11/23, 110 YDS` → `11/23 completions for 110 passing yards`
- `10 CAR, 39 YDS, 1 TD` → `10 carries for 39 rushing yards and 1 rushing touchdown`
- `2 REC, 46 YDS` → `2 receptions for 46 receiving yards`
- Team rushing context now says `including 85 rushing yards per game`

The behavior is covered by a regression fixture based on the Steelers–Bills payload. This is a constrained NFL wording correction and does not expand the formatter’s sport contract.

egyptiangio:
Thanks for the thorough follow-ups — every term from the last round is met, the freeze commit is clean, and the trial merge against current dev is green (the conflict with #681's overlay change resolves trivially; both sides add an `elif` and they compose).

One final structural ask before merge, and it's about the only part of this PR I'd otherwise be maintaining by hand forever: **the preview field list is currently hand-typed in four places** — `_GENERATED_PREVIEW_FIELDS` in `sports_data.py`, `_DETAIL_FIELDS` in `generated_preview.py`, the 40 kwargs in `team_matcher.py`'s cache rehydration, and the serializer in `provider_cache.py`. That's the exact pattern that caused the silent field-drop your own `test_match_cache_short_name_heal` fixes: a field added to the dataclass but missed at one site just quietly renders empty on the cache-hit path.

Please consolidate to **one canonical tuple** (e.g. `GENERATED_PREVIEW_FIELDS` in `core/types.py`, next to the dataclass) and have the other sites iterate it dict-style (`Event(**{f: cached_data.get(f, "") for f in GENERATED_PREVIEW_FIELDS}, ...)` and the like) instead of listing names. Then add a parity test in the spirit of `tests/test_settings_registry.py`: every field in the tuple (a) exists on `Event`, (b) survives a `provider_cache` serialize→rehydrate round trip, (c) is covered by the refresh-overlay snapshot rule, and (d) has a registered template variable. The renderer's `_DETAIL_FIELDS`/`_ENRICHED_FIELDS` can derive from the canonical tuple too.

No behavior change intended — this just makes "add field #41" a one-line edit with CI enforcement instead of a four-file memory exercise. With that in, I'm happy to merge under the arrangement we agreed.

ncsufan8628:
Updated as requested.

- Rebased the PR branch onto the latest `dev`.
- Added a single canonical `GENERATED_PREVIEW_FIELDS` tuple in `teamarr/core/types.py`.
- Updated `sports_data`, `generated_preview`, match-cache rehydration, and provider-cache serialization/deserialization to derive their generated-preview fields from that tuple.
- Added parity coverage proving that every canonical field:
  - exists on `Event`;
  - survives a provider-cache round trip;
  - participates in the post-kickoff snapshot-preservation rule; and
  - has a registered public template variable.
- Preserved backward compatibility for older cache entries, including the existing `week=None` default.
- Kept the pregame snapshot-preservation fix in its own commit.
- While resolving the rebase conflict in `sports_data.py`, retained both the generated-preview snapshot behavior and the zero-score handling added by #681.

The consolidation itself does not change generated-preview behavior; it removes the duplicated field lists and makes omissions detectable through one parity test.

Validation completed:

- `2770 passed`
- Ruff: passed
- Pyright: 0 errors
- `git diff --check`: passed

PR #589, opened 2026-08-22
fix(tsdb): handle local dates across UTC boundaries

### LINKED ISSUE #588: fix(tsdb): match events by dateEventLocal across UTC date boundaries
## Problem

TheSportsDB events can have different UTC and venue-local calendar dates. Teamarr currently filters some TSDB fallback responses using only `dateEvent`, which can exclude an event from the date users see locally.

This affects sports generally, but it was observed while matching a Boxing event.

## Example

A Boxing source was generating event-card streams such as:

```text
LIVE EVENT 02 - 6/8pm Rolly v Teofimo
LIVE EVENT 03 - (HD) Rolly v Teofimo
```

Teamarr correctly classified the streams:

```text
[CLASSIFY] 'LIVE EVENT 02 - 6/8pm Rolly v Teofimo' -> event_card
[CLASSIFY] 'LIVE EVENT 03 - (HD) Rolly v Teofimo' -> event_card
[CACHE_HIT] events:boxing:2026-08-22
```

However, neither stream matched an event for August 22.

TheSportsDB event `2528767` contains:

```json
{
  "idEvent": "2528767",
  "strEvent": "Rolando Romero vs Teofimo Lopez",
  "dateEvent": "2026-08-23",
  "dateEventLocal": "2026-08-22",
  "strTimestamp": "2026-08-23T01:00:00",
  "strTime": "01:00:00",
  "strTimeLocal": "18:00:00"
}
```

The event occurs on August 22 at its venue, but its UTC calendar date is August 23.

Changing the manual-match date to August 23 makes the event appear, even though Teamarr displays its converted time as August 22 for an `America/New_York` user.

## Cause

In `teamarr/providers/tsdb/provider.py`, the `eventsnextleague` fallback filters raw events using only `dateEvent` before parsing the event timestamp:

```python
event_date = event_data.get("dateEvent")
if event_date != date_str:
    continue
```

The season fallback and team-schedule season fallback contain equivalent filtering.

As a result, an event is discarded before Teamarr can parse its UTC timestamp and convert it into the configured user timezone.

## Proposed behavior

When filtering TSDB events for a requested date, accept the event when either `dateEvent` or `dateEventLocal` matches:

```python
event_date = event_data.get("dateEvent")
event_date_local = event_data.get("dateEventLocal")

if date_str not in {event_date, event_date_local}:
    continue
```

This should be implemented through a shared helper and applied consistently to:

- `eventsnextleague` fallback filtering
- `eventsseason` fallback filtering
- Team-schedule season fallback filtering
- Any other TSDB path that filters raw events by `dateEvent`

Events without `dateEventLocal` should preserve the existing behavior.

## Expected result

For an `America/New_York` user requesting Boxing events for August 22:

- Event `2528767` is retained because `dateEventLocal` is `2026-08-22`.
- Its original timestamp remains `2026-08-23T01:00:00Z`.
- Timezone conversion places it on August 22 locally.
- The event reaches the event-card matcher.
- `Rolly v Teofimo` can fuzzy-match `Rolando Romero vs Teofimo Lopez`.

The fix should remain general: no event-specific fighter keywords or manual time correction should be introduced.

ncsufan8628:
## Summary

- accept a TSDB event when either `dateEvent` or `dateEventLocal` matches the requested calendar date
- apply the shared date check to next-event, season, and team-schedule fallbacks
- preserve the original TSDB timestamp and existing behavior when `dateEventLocal` is absent
- version the persisted event-cache namespace so pre-fix empty results cannot mask newly discoverable events after upgrade
- cover the UTC/local boundary through the real Boxing event-card classifier and matcher

No fighter-specific keywords or time corrections are introduced.

Fixes #588

## Testing

- `pytest -q` — 2252 passed
- `ruff check teamarr/providers/tsdb/provider.py teamarr/services/sports_data.py tests/providers/test_tsdb.py tests/matching/test_event_card_fuzzy_match.py`

### DIFF
--- a/teamarr/database/schema.sql
+++ b/teamarr/database/schema.sql
@@ -1460,7 +1460,7 @@ INSERT OR IGNORE INTO cache_meta (id) VALUES (1);
 -- =============================================================================
 
 CREATE TABLE IF NOT EXISTS service_cache (
-    -- Cache key (e.g., "events:nfl:2026-01-06")
+    -- Cache key (e.g., "events_v2:nfl:2026-01-06")
     cache_key TEXT PRIMARY KEY,
 
     -- Cached data (JSON serialized)
--- a/teamarr/providers/tsdb/provider.py
+++ b/teamarr/providers/tsdb/provider.py
@@ -134,9 +134,11 @@ class TSDBProvider(SportsProvider):
         if data and data.get("events"):
             events = []
             for event_data in data["events"]:
-                # Filter to target date
-                event_date = event_data.get("dateEvent")
-                if event_date != date_str:
+                # TSDB's canonical date is UTC for some events, while
+                # dateEventLocal is the venue-local calendar date.  Accept
+                # either so events crossing midnight UTC remain discoverable
+                # from the date users see on the schedule.
+                if not self._event_matches_date(event_data, date_str):
                     continue
                 event = self._parse_event(event_data, league)
                 if event:
@@ -153,9 +155,7 @@ class TSDBProvider(SportsProvider):
         if data and data.get("events"):
             events = []
             for event_data in data["events"]:
-                # Filter to target date
-                event_date = event_data.get("dateEvent")
-                if event_date != date_str:
+                if not self._event_matches_date(event_data, date_str):
                     continue
                 event = self._parse_event(event_data, league)
                 if event:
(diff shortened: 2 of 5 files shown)


egyptiangio:
Thanks for this — the report in #588 was exactly right, and your fix is correct for the reported case. Reviewing it surfaced that the same date-boundary bug exists in six more providers, so we're fixing the class structurally instead: #590 / #592 moves date membership to the service layer entirely (providers now return supersets; the user's calendar decides). Your boxing boundary scenario is pinned as a regression test in the new suite. Closing in favor of that branch — the changelog will credit you and #588 for the find.

Issue #590, opened 2026-08-23
Calendar dates shouldn't cross the provider boundary — filter events by user-day window at the service seam


egyptiangio:
**Problem**

`get_events(league, target_date)` receives a user-local calendar date, but each provider answers "is this event on that date?" against someone else's calendar — raw provider date strings or UTC `.date()`. Every such comparison disagrees with the user's calendar near midnight boundaries.

This class of bug has been fixed locally three times (#345 UFC cards spanning midnight, #588/#589 TSDB UTC vs venue-local dates). An audit finds 6 more latent sites:

| Site | Compares |
|---|---|
| Squiggle AFL (`_game_on_date`) | UTC date, documented as such — worst case, Australia is UTC+10/11 |
| NASCAR (`provider.py:121`) | UTC session date |
| TSDB racing sessions (`provider.py:117`) | UTC session date |
| Bell Media (`provider.py:48`) | raw `date` string |
| HockeyTech (`client.py:226`) | raw `date_played` string |
| Supabase (`client.py:492`) | raw `game_date_override` string |

(+ the three TSDB fallback paths #588 reported.) The class recurs because every provider re-implements the date decision.

**Fix (structural)**

1. `user_day_window(date) → (start_utc, end_utc)` — the only place user tz is consulted; DST-safe.
2. The service layer (`sports_data.get_events`) filters all provider results through `event_intersects(event, window)` — pure instant math, single enforcement point. Providers cannot forget the filter because they no longer own it.
3. Provider contract becomes "return a superset": raw prefilters widen to ±1 day; exactness is no longer their job. New providers are correct by construction.
4. Events cache: namespace bump + tz-aware keying (also fixes the latent "changing display timezone poisons cached per-date results").

**Known edge, explicitly open:** a server-side day-bucketed fetch (ESPN scoreboard `?dates=`, MLB Stats schedule) can still under-return at day edges for users far from the API's home calendar; widening those fetches is a contained follow-up if ever observed — the seam makes the gap visible in one place instead of nine.

Supersedes PR #589 (thanks @ncsufan8628 — the report and regression scenarios from #588 carry into the new tests); fixes #588.

egyptiangio:
Landed on dev in 2469865f (PR #592). Closes at next release.

egyptiangio:
Released in [v2.13.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.13.0).

PR #602, opened 2026-08-25
fix: union provider day buckets at the date seam (#601)

### LINKED ISSUE #601: [Bug]: Upgrading from 2.12.2 to 2.13.0 Seams to have broken channel mapping of games within lookup days
### Tag / Commit

latest

### Dispatcharr version

0.29

### Browser

_No response_

### Affected area

None

### What happened?

Since updating to v2.13 i dont appear to have streams from Serie A Italian legue or champions league qualifying anymore, I see the streams in source preview. I checked this on 2 teamarr instanes on my test windows docker system and my ubuntu instance, On both systems reverting back to 2.12.2 fixed the issues.

### Expected behavior

Streams should have mapped as they were from 2.12.2

### Steps to reproduce

Update from 2.12.2 to 2.13.0

### Logs

```shell

```

### Screenshots

[Log Teamarr.txt](https://github.com/user-attachments/files/31422169/Log.Teamarr.txt)

lpukatch:
Fixes #601.

## The bug

Serie A and Champions League qualifying vanished entirely after 2.13.0 for users far from the API's home calendar.

#590 moved date membership to a single seam: `get_events` filters everything providers return through the **user-local** day window. But the *fetch* stayed bucketed by the **provider's** calendar — ESPN's scoreboard `?dates=` and MLB Stats file each event under their own day and structurally cannot honour the "return a ±1-day superset" contract the seam assumes.

For evening European kickoffs the two calendars disagree, and the event falls through both ends:

- filtered out of day `D` — past the end of the local window
- never fetched under day `D+1` — the provider files it under `D`

So it is invisible at **every** lookahead value, which is why the reporter raising lookahead to 7 days and the buffer to 168h changed nothing.

The #590 commit message flagged this as a known open edge. What wasn't appreciated is that it deletes entire leagues, not just day-boundary strays.

## Measured against live ESPN data

Serie A, next 7 days, identical code, only the user timezone varying:

| User TZ | ESPN events | Visible to matcher | Lost |
|---|---|---|---|
| America/New_York | 10 | 10 | 0 |
| Europe/London | 10 | 10 | 0 |
| Australia/Sydney | 10 | **0** | **10** |

`eng.1` for Australia/Sydney was **4 visible / 6 lost** — the 11:30Z and 13:00Z kickoffs survive, everything from 14:00Z on does not. That is exactly why the reporter still saw some Premier League matches (their log shows 8) and zero Serie A / UCL qual: Serie A is all 16:30Z/18:45Z and UCL qual all 19:00Z/20:00Z.

The cutoff for AEST (+10) is 14:00Z. Generally, for offset +N, provider-day-`D` events at or after `(24−N):00Z` are unreachable.

**After the fix**, same live data and timezone: `ita.1` 12 visible, `uefa.champions_qual` 7, `eng.1` 11 — each landing on the correct local day.

## The fix

`get_events` unions the provider-day buckets `D-1, D, D+1` (`provider_day_buckets`) and dedupes by `(provider, event id)` before applying the window filter. The union has to happen on the fetch side; asking providers to over-return can't work for server-side bucketing.

- **Cost**: raw buckets get their own timezone-independent cache layer (`events_raw:<provider>:<league>:<date>`). Consecutive target dates share buckets, so a contiguous run of N days costs **N+2** provider calls, not 3N.
- **Isolation**: a failing neighbour bucket is logged and skipped rather than costing the day actually requested.

## Tests

- New coverage for the day-bucketed case: the AEST regression itself, bucket fan-out, cross-date bucket reuse, tz-independence of the raw layer, dedup, and neighbour-failure isolation.
- Call-budget tests re-baselined with the constants explained. The invariants they protect all still hold: repeats stay flat, neighbouring dates share buckets, cost stays independent of team count.
- Fixed the seam tests' timezone fixture, which patched a re-export neither module binds — those tests were silently reading the host machine's timezone and passed only because it happened to be `America/New_York`.

## Gates

`ruff` clean · `2301 passed` · `npm run build` clean · docs updated (`consumer-layer.md`)

Bead: `teamarrv2-z9y2`


egyptiangio (commented):
Reviewed, and I reproduced both the bug and the fix locally rather than going off the numbers in the description. **This looks right to me — good diagnosis and the correct minimal shape.** Findings below are yours to take or leave; only the first one would I consider doing before merge, and it's a one-liner.

## Reproduced independently

One note for anyone else verifying: my first attempt showed *no* bug, because the persistent cache singleton served warm data from the previous run. With an isolated cache it reproduces exactly as you describe:

```
ON DEV (pre-fix)          ita.1   ucl_qual   eng.1
  America/New_York          10        4        10
  Australia/Sydney           0        0         4     <- entire leagues gone

ON PR #602 (fixed)
  America/New_York          10        4        10     <- home region untouched
  Australia/Sydney          10        7        14
```

I also confirmed the assumption the fix rests on but doesn't state: **ESPN buckets by US Eastern, not UTC** — a 01:40Z game files under the previous day's `?dates=`. That's what makes D-1..D+1 sufficient. See finding 3.

Gates on the branch here: `ruff` clean, **2301 passed**.

## 1. The raw bucket layer isn't single-flighted (worth fixing before merge)

`_fetch_provider_day` does get -> fetch -> set with no `lock_key`. The outer lock is on `events_v2:<league>:<date>:<tz>`, so two concurrent requests for *different target dates* hold *different* locks while sharing buckets. With a real lock-bearing cache and two concurrent target dates:

```
provider calls = 6   (sequential ideal = 4: buckets 28,29,30,31)
```

Buckets 29 and 30 each got fetched twice. Not a correctness problem, but it erodes the singleflight guarantee `test_cache_singleflight.py` exists to protect, and it'd show up as duplicate scoreboard requests during parallel stream matching. Your new `== 4` assertion holds sequentially but isn't guaranteed under concurrency.

`with self._cache.lock_key(raw_key):` around the raw fetch should close it.

## 2. ESPN supports date ranges — follow-up, not this PR

```
?dates=20260828-20260830  ->  8 events, all three days, ONE request
```

Provider-agnostic is the right call for the fix, and the amortization argument holds. But the 3x cost is real on a cold cache for isolated lookups (`get_sample_event`, single-day refresh), and collapsing it for the dominant provider would be a clean follow-up if you think it's worth a bead.

## 3. The margin is tighter than the docstring implies

"One day either side covers every real UTC offset (max ±14h) plus `EVENT_TAIL_HOURS` of spill" undersells how close it runs. The real bound is `offset_user − offset_provider <= 21h` (24h span minus the 3h tail). ESPN at −4 with a UTC+14 user is 18 <= 21 — safe, but only 3h of headroom, and it hinges on a provider bucketing offset that isn't written down anywhere. Putting the inequality in the docstring would make a future provider at a more extreme offset trip a comment rather than a silent hole.

## 4. TSDB pre-warm now under-warms

`sports_data.py:935` still warms only `target_date`, not D-1/D+1, so it no longer covers what the seam asks for. Cosmetic — the missing buckets just fetch on demand.

I checked for correctness bypasses and found none; the other `provider.get_events` hits are a docstring example and provider-internal calls.

---

Separately: **the timezone-fixture fix is the most valuable thing in here after the fix itself.** Those seam tests were patching a re-export neither module binds, so they were silently reading the host machine's timezone and passing only because the box happens to be `America/New_York`. That's precisely why #590's own tests didn't catch #590's bug — and it means any seam test written before this PR proved less than it looked like it did. Good catch.

Issue #605, opened 2026-08-25
Generation performance: hot-path optimization pass (matching, cache eviction, event prefetch, template resolution)


lpukatch:
## Summary

A profiling pass over EPG generation found four independent hot paths doing large amounts of redundant work. All four are fixed on `perf/generation-hot-paths`; this issue tracks the work through review and release.

## Measured results

Benchmarked against the committed matching corpus (400 streams × 3047 events across 10 leagues) on the same machine, before vs. after:

| Hot path | Before | After |
|---|---|---|
| `match_multi_league` | 48.8 ms/stream | **13.9 ms/stream** (3.5×) |
| `TTLCache.set` of a new key into a 45k-entry cache | 1.93 s / 2000 inserts | **0.0019 s / 2000** (~1000×) |
| Template resolution, per programme (5 fields) | 0.52 ms | **0.12 ms** (4.4×) |
| Event prefetch, network-bound portion | serial | **5–10×** (fake service at 10 ms/call) |

Extrapolated, the stream-matching phase for 5000 streams drops from roughly 245 s to roughly 70 s.

## What was wrong

**1. Match loop — `matching/team_matcher.py`, `matching/identity.py`**

- `_check_alias_match` generated both event teams' pattern lists *before* checking whether either stream side resolves to an alias. Aliases are rare, but this runs once per (stream × candidate event), so every alias-less stream paid two list allocations per candidate — the single largest wasted allocation in the loop.
- Pure per-pair kernels were recomputed from scratch every time: `_best_name_score`, `_short_name_leg_is_safe`, the abbreviation path's stream-side tokens, `residual_contradicts`' token sets, and the per-`(name, league)` alias lookup. `_best_name_score` now runs at an 86% hit rate, because a team plays several games inside the match window.
- Two `astimezone` conversions per (stream × event) for values that depend only on the event.
- The flattened `[(league, event)]` candidate list was rebuilt from the same prefetch for every stream — tens of thousands of tuple allocations each.

**2. `utilities/cache.py` — `TTLCache` eviction**

`_evict_if_needed` swept the entire cache for expired entries on every insert of a new key, even when nowhere near `max_size`. At 45k entries that is roughly 1 ms per `set`, paid thousands of times per generation run. Separately, LRU eviction was a full `min()` scan *per victim*.

**3. `matching/matcher.py` — event prefetch**

[...]

Issue #599, opened 2026-08-25
[Bug]: Cross-sport streams attach to unrelated channels when team names share a city


egyptiangio:
Reported on Discord (2026-08-24) with logs. Streams naming teams from a different sport are being attached to channels they have nothing to do with, whenever the two names happen to share a city.

## Reported behaviour

An MLB-scoped `ESPN + 2` source created the right channels, then attached the wrong streams:

```
[RESOLVER] League mode: mlb -> 'MLB' -> group_id=5220
[CREATED] Managed channel id=2 name=MLB | TB/DET event=401816657
[CHANNEL_CREATE] id=7815 (#700) stream='ESPN+ 81 (D): Tampa Bay Lightning vs. De' event=401816657
[STREAM_AUDIT] consolidate add: ch='MLB | TB/DET' (db_id=2, d_id=7815) added stream_id=79011
```

and

```
[CREATED] Managed channel id=3 name=MLB | COL/WSH event=401816656
[CHANNEL_CREATE] id=7816 (#701) stream='ESPN+ 146 (D): Northern Colorado vs. Eas' event=401816656
```

So an **NHL** game (Lightning vs Red Wings) landed on the MLB Rays/Tigers channel, and an **NCAA** game (Northern Colorado vs Eastern Washington) landed on the MLB Rockies/Nationals channel.

## Root cause

`token_set_ratio` weighs every token equally, so a shared **city** — the least informative half of a team name — is enough on its own:

| stream side | event team | score |
|---|---|---|
| Tampa Bay Lightning | Tampa Bay Rays | 78.3 |
| Detroit Red Wings | Detroit Tigers | 71.0 |
| Northern Colorado | Colorado Rockies | 66.7 |
| Eastern Washington | Washington Nationals | 73.7 |

The both-teams path accepts when the *weaker* leg clears `BOTH_TEAMS_THRESHOLD = 60` (`team_matcher.py:1616`), so two weak legs prop each other up and the match is accepted at 71% and 67% "confidence".

Nothing downstream rejects it. `classify_stream` returns `sport_hint=None` and `league_hint=None` for these names, so the sport gate (`team_matcher.py:979`) never fires; the trusted-date gate (`:1005`) needs a date in the stream name and there isn't one. The candidate window is −30/+3 days, so ~500 MLB games are eligible and the best-scoring one wins.

The engine never asks the question a human would ask first: **do these two teams actually play each other?**

## Scope

[...]

PR #606, opened 2026-08-25
perf: generation hot-path optimization pass (#605)

lpukatch:
Closes the work tracked in #605. Bead: `teamarrv2-2nmp`.

**Leaving this open for maintainer feedback rather than merging** — it touches the matching inner loop and the shared TTL cache, and a second pair of eyes on the invariants below is worth more than a fast merge.

## What this is

A profiling pass over EPG generation. Four independent hot paths were doing large amounts of redundant work. The changes are **reordering and memoization only** — no matching semantics, thresholds, ranking rules, or template behavior were changed.

## Measured

Benchmarked against the committed matching corpus (400 streams × 3047 events across 10 leagues), same machine, `dev` vs. this branch:

| Hot path | Before | After |
|---|---|---|
| `match_multi_league` | 48.8 ms/stream | **13.9 ms/stream** (3.5×) |
| `TTLCache.set` of a new key into a 45k-entry cache | 1.93 s / 2000 inserts | **0.0019 s / 2000** (~1000×) |
| Template resolution, per programme (5 fields) | 0.52 ms | **0.12 ms** (4.4×) |
| Event prefetch, network-bound portion | serial | **5–10×** (fake service at 10 ms/call) |

Extrapolated, the stream-matching phase for 5000 streams drops from roughly 245 s to roughly 70 s. The prefetch figure used a 10 ms fake latency; real scoreboard calls are 100–500 ms, so that one should be larger in practice.

## The four changes

**1. Match loop** — `matching/team_matcher.py`, `matching/identity.py`

`_check_alias_match` generated both event teams' pattern lists *before* checking whether either stream side resolves to an alias. Aliases are rare, but this runs once per (stream × candidate event), so every alias-less stream paid two list allocations per candidate. Now it resolves first and bails early — the both-sides-required return condition means a half-resolved pair can bail just as early as an unresolved one.

The pure per-pair kernels are now memoized: `_best_name_score` (86% hit rate — a team plays several games inside the match window), `_short_name_leg_is_safe`, the abbreviation path's stream-side tokens, `residual_contradicts`' token sets, and the per-`(name, league)` alias lookup (cleared by `reload_aliases`, so a user's edit still takes effect immediately). `_local_date` replaces the two `astimezone` calls per (stream × event) — kick-off times cluster on the hour, so distinct keys are far fewer than events.

The flattened `[(league, event)]` candidate list was rebuilt from the same prefetch for every stream. It is now built once per (league set, date window); the returned list is shared and read-only, and every remaining `all_events.append` is on the non-prefetched fallback path's own fresh list.

**2. `TTLCache` eviction** — `utilities/cache.py`

`_evict_if_needed` swept the entire cache for expired entries on every insert of a new key, even with room to spare — roughly 1 ms per `set` at 45k entries, paid thousands of times per generation run. The sweep is now gated on actually being full, and runs on an interval once there. LRU eviction moved to `OrderedDict`: O(1) per victim, where it was a full `min()` scan per victim. Eviction order is exact, and `max_size` is still never exceeded.

**3. Event prefetch** — `matching/matcher.py`

`_prefetch_events` now plans every (league × date) cell, fills the cells that still need a service call, then assembles in league order. Network-bound fetches go out concurrently (`ESPN_MAX_WORKERS`, default 24 here).

Two things stay on the calling thread on purpose: cache-only cells never touch the network, so a thread would only add GIL hand-off to pure deserialization; and TSDB's rate limiter sleeps *while holding its lock*, so concurrent callers would queue inside that sleep rather than overlap. Both are asserted by tests.

**4. Template resolution** — `templates/resolver.py`, `consumers/team_epg.py`, `consumers/event_epg.py`

`resolve()` runs all 252 registered extractors plus their `.next`/`.last` suffixes on every call, and a single programme resolves five to eight fields against a context that does not change between them. `build_variables()` lets a caller build the map once and pass it as `variables=`.

Deliberately a **returned value rather than a cache on the resolver**: one `TemplateResolver` instance is shared across the team scan's worker threads, so resolver-held state would be a data race. The two call sites that mutate their context (`feed_team`, `extra_vars`) build the map after mutating.

Two smaller ones: `LeagueMappingService.get_mapping_by_league` linear-scanned all mappings on every provider probe of every league (now an O(1) index, built from `_mappings` after load so it resolves exactly as the scan did); and full EPG generation no longer merges the group guide into a `total_xmltv` it then discards (`aggregate_xmltv=False`; the flag defaults to `True`, so `GroupService.process_all_groups` is unaffected).

## Verification

- 2322 tests pass (20 new), `ruff check` clean, `npm run build` clean.
- `tests/matching/test_prefetch_parallel.py` pins the fan-out against a **verbatim port of the old serial loop**, asserting the same per-league events, the same `shared_events` state, and the same *set of service calls*, across four scenarios (cold, warm `shared_events`, all-subscribed, none-subscribed). It also asserts TSDB never leaves the calling thread and that one league's outage does not cost the batch.
- `tests/test_cache_eviction.py` pins never-exceeds-`max_size`, LRU order, expired-first reclamation, and — timed, because the defect was purely asymptotic — that inserting into a roomy cache does not scale with cache size.
- `tests/templates/test_shared_variable_map.py` pins that a pre-built map resolves identically to rebuilding per field, including unknown variables, contextless suffixes, filters, and `extra_vars`.

## What I'd most like reviewed

1. **Prefetch concurrency default.** `PREFETCH_MAX_WORKERS` reads `ESPN_MAX_WORKERS` (default 24 here). If a user has set that to 100 for the team scan, this path inherits 100 concurrent scoreboard requests. Sharing the knob felt right; a separate variable or a hard cap is equally defensible.
2. **`_best_name_score_cached` memory.** `lru_cache(maxsize=65536)` over 4-string keys, process-lifetime. Bounded, but it is the largest new allocation.
3. **The shared candidate list** is returned read-only by convention rather than as a tuple. I checked every consumer; a `tuple` return would enforce it at some allocation cost.

## Found but deliberately not changed

- `StreamMatchCache.touch()` commits per stream, contradicting `session()`'s own documented "commits once at session end" contract. Measured at ~8 µs/commit on SSD (negligible), but on a NAS or spinning-disk Docker volume it would be seconds per group.
- SQLite runs at `synchronous=FULL` (the WAL default). `NORMAL` roughly halved commit cost in measurement, but it changes crash durability — a maintainer call, not a mechanical optimization.
- Event groups are still processed sequentially under one shared connection. Parallelizing them looked risky around channel-number allocation ordering, and `shared_events` already makes later groups' prefetch nearly free.

## Docs

Updated for the third `ESPN_MAX_WORKERS` default (`configuration.md`, `providers/espn.md`) and the prefetch architecture (`architecture/consumer-layer.md`).


lpukatch:
Bead: `teamarrv2-2nmp` (in_progress — will close when this merges). Issue: #605.

egyptiangio:
## Review — independent second pass

Verified cold against `dev...HEAD`: ruff clean, 2322 tests pass, and I re-ran the TTLCache benchmark independently (2.12s → 0.0015s for 2k inserts at 45k entries — the ~1000x claim is real). The "reordering and memoization only" claim holds almost everywhere: I checked every memo for purity and invalidation (`_local_date`, `_code_tokens`, `_short_name_leg_is_safe`, `_best_name_score_cached`, alias memo vs `reload_aliases`, candidates memo, league-mappings index), truth-tabled the `_check_alias_match` reorder, confirmed no context mutation between `build_variables` and its consumers, and confirmed the parallel prefetch sits on the pre-existing single-flight locks (`events_v2` → `events_raw`, no reverse path, no cycle). Solid work.

**One blocking issue — the prefetch error path changes semantics in the silent-missing-match direction (the #599/#601 failure mode):**

1. **Failed slots poison `shared_events`.** When a concurrent `get_events` raises, the slot gets `events=[]` and pass 3 publishes `([], cache_only=False)` under that key. Later groups' reuse check (`shared_events or not was_cache_only or ...`) accepts empty-but-fresh, so one exception silently blanks that (league, date) for the rest of the run. Previously this failed loudly per-group (processor.py:873) and later groups re-fetched. Fix: mark failed slots and skip publishing them.
2. **Inline slots (TSDB + cache-only) have no try/except at all** — a TSDB exception (the rate-limited, flakiest path) still kills the whole prefetch. Make it consistent with the concurrent path.
3. **Pin both with tests.** `test_a_failing_league_does_not_lose_the_batch` only asserts healthy leagues survive; it never inspects `shared_events` and never exercises a TSDB failure.

**Non-blocking, worth doing in this PR (cheap):**

4. `test_expired_entries_are_reclaimed_before_live_ones` passes only because `max_size=10` degenerates the sweep interval to 1. At production size (50k → sweep every 2500th at-capacity insert) the named property is false. Rename/reframe it so the suite doesn't pin a property the code doesn't have.
5. Nothing calls `cleanup_expired()`, and `_evict_if_needed` now returns early below capacity — expired entries are never proactively reaped until the cache is full (previously every new-key insert swept them). Fine for correctness; either wire `cleanup_expired()` to a periodic task or note the memory tradeoff in the docstring.
6. `_prefetched_candidates` returns a shared mutable list guarded only by a comment, and the memo's freshness rests on the prefetch dict never being mutated in place. Return a tuple and add an assert/comment at the producer so a future refactor fails loudly instead of losing matches quietly.

**Soak, don't block:** the interaction with #602 is real but untested against a live provider. Per-bucket caching means there's no serial "3x fan-out" penalty, and across leagues buckets share nothing — so effective ESPN concurrency during prefetch is essentially the full worker count (default 24). A rate-limit burst during prefetch gets silently cached as empty days by pre-existing service-level behavior; this PR makes that much easier to trigger. Recommend a dev soak with logs watched for `[PREFETCH] ... failed` / `[EVENTS] ... bucket ... failed` before release, and consider a lower default if 429s show up.

**Path to merge:** fix 1–3 (one small commit: `failed` flag on `_PrefetchSlot`, skip publish, symmetric try/except, two test assertions), ideally fold in 4–6, then merge and soak on dev per above. Everything else verified good to go.

lpukatch:
Thanks — the blocking catch is right, and it's the one that mattered. Publishing `([], cache_only=False)` for a failed slot is strictly worse than what it replaced: the reuse check treats empty-but-freshly-fetched as authoritative, so a single exception blanked that (league, date) for the whole run, where the serial version failed loudly per group and let the next group re-fetch. That's the #599/#601 direction and I should have caught it when I wrote the `except`.

All six addressed in two commits.

## Blocking (1–3) — `984cd66b`

**`_PrefetchSlot` gains `failed`.** Pass 3 skips publishing those slots, so the key stays absent and the next group issues a real fetch. The docstring now says why an empty `events` on a failed slot is an absence of knowledge rather than an answer.

**Inline slots get the same `try`/`except`.** Both paths now go through one `_record` helper, so there's a single place the policy lives. You're right that TSDB was the worst gap — flakiest call we make, and it had no isolation at all.

**A WARNING summary names the affected leagues**, which is the line to watch during the soak:

```
[PREFETCH] 5/408 fetches failed and were NOT cached; affected leagues: lg1
```

**Four tests, and I verified each fails without its fix.** Reverting the `not slot.failed` guard fails 3; removing the inline `try`/`except` fails 2.

One correction to the test you'd have written from the description: asserting *nothing* under `lg1:` gets published is too strong. The cache-only slots for a failed league don't fail — they legitimately return empty from cache, and `([], cache_only=True)` is correctly rejected by the reuse check for any group that needs the league. So the assertion is on the exact poisoning condition, `empty AND not cache_only`. There's also a behavioural test that a later group really does re-fetch the failed league, which is the property that actually matters.

## Non-blocking (4–6) — `66849601`

**4.** Correct, and a good catch — `max_size=10` puts the sweep interval at 1, so the test proved nothing about production. Split into two honest tests: one pins the behaviour on a sweep tick (and asserts the interval *is* 1, so the setup can't drift silently), the other pins what actually happens between ticks — plain LRU, where an expired-but-recent entry outlives a live-but-old one.

**5.** Wired `cleanup_expired()` into `PersistentTTLCache._background_flush`, which already runs on a timer. Measured at ~2.4 ms per 50k entries, once per flush interval. Reap, flush and reschedule are isolated from each other so no failure skips the next step (tested). The residual trade-off is documented on `_evict_if_needed` for the standalone `TTLCache` case, which has no timer.

**6.** `_prefetched_candidates` returns a tuple now; consumers widened to `Sequence` (they only iterate), and the non-prefetch fallback paths still build and append to their own list — pyright confirms the narrowing. I also added the half of the invariant you identified but that had nothing enforcing it: a note at `_prefetch_events` that it must *replace* `self._prefetched_events` rather than mutate it, since the memo keys on that dict's identity, plus a test that a new prefetch invalidates the memo.

## Verification

`ruff` clean, pyright `0 errors`, **2331 tests pass** (+9). Re-ran both benchmarks: match loop unchanged at 14.3 ms/stream, and the serial-equivalence harness still reports identical events, identical `shared_events` and an identical set of service calls across all four scenarios.

## On the soak

Agreed, and your read of the #602 interaction matches mine — per-bucket caching means no serial fan-out penalty, but it also means effective concurrency during prefetch is the full worker count. The `[PREFETCH] N/M fetches failed` summary should make a 429 burst obvious rather than something you have to infer from missing channels. Two things worth noting for whoever watches it:

- A rate-limited fetch that *raises* is now visible and uncached. A rate-limited fetch that returns an empty day still caches as empty — that's pre-existing service-level behaviour this PR doesn't change, and it's the harder-to-see half.
- If 429s do show up, `PREFETCH_MAX_WORKERS` is a one-line default and already overridable via `ESPN_MAX_WORKERS` without a release.

Happy to lower the default to something more conservative pre-merge if you'd rather soak from the cautious end — say the word and I'll push it.

PR #608, opened 2026-08-25
fix: run enforcement outside the group processor's write transaction (#607)

### LINKED ISSUE #607: Enforcement deadlocks against the group processor's own uncommitted transaction (30s stall + dropped writes every run)
Found while watching a live generation on Kubernetes during the #606 soak. **Not caused by #606** — see Provenance below.

## Symptom

Every generation run stalls for exactly 30 seconds and then drops keyword-enforcement writes:

```
13:32:16 | INFO  | ...stream_fetcher:380 | [FILTER] Group 'Live | Ligue1 🇫🇷': 5 input → 0 passed
13:32:46 | ERROR | ...enforcement.keywords:255 | [KEYWORD_ERROR] database is locked
Traceback (most recent call last):
  File "/app/teamarr/consumers/enforcement/keywords.py", line 139, in enforce
    remove_stream_from_channel(
  File "/app/teamarr/database/channels/streams.py", line 102, in remove_stream_from_channel
    cursor = conn.execute(
sqlite3.OperationalError: database is locked
```

30 seconds is exactly `PRAGMA busy_timeout=30000` (`database/connection.py:60`). The step never succeeds — it waits out the full timeout and gives up.

Observed **11 times today**, on every run since 05:35. On the run I measured it was **31s of a 220s groups phase — 14%**.

## Mechanism

It is a self-deadlock between two connections on the same thread.

1. `EventGroupProcessor.process_all_groups` opens a connection at `processor.py:292` and holds it for the entire groups phase.
2. That connection **writes** — `store_group_xmltv(conn, ...)` runs per group (`processor.py:624`, `processor.py:845`). SQLite takes the WAL write lock at the first write and holds it until commit.
3. There is **no `conn.commit()` anywhere in `processor.py`** — the commit happens when `get_db()`'s context manager exits, i.e. after the whole phase.
4. Enforcement runs *inside* that block (`run_step("keyword", ...)`, `processor.py:462`), and `KeywordEnforcer.enforce()` opens its **own** connection at `keywords.py:99`.
5. That second connection tries to `UPDATE managed_channel_streams` → blocks on the write lock → the lock cannot be released until the outer block exits → the outer block cannot exit until enforcement returns.

No timeout can resolve this. It fails 100% of the time, whenever the step actually has a write to perform.

## Blast radius is wider than keywords

The same pattern — open a fresh connection inside the outer block — appears in:

| Site | Writes? |
|---|---|
| `enforcement/keywords.py:99` | yes — **currently failing** |
| `enforcement/ordering.py:87` | yes |
| `lifecycle/cleanup.py:132, 337, 607, 702` | yes |

Only the step that actually has work to do trips it, which is why this stayed hidden. In the run I captured, `cleanup_disabled_groups` ran immediately after the failure and logged `Checking 1 disabled group(s)` with no error — it found nothing to delete, so it never attempted a write. Keyword enforcement is simply the step that currently has work.

That also explains the timing: the bug is likely months old and became *reachable* today when a data condition changed (a keyword whose behavior is `ignore` started matching a live stream). The errors log goes back to 2026-04-03 with no prior occurrence.

## Probable second bug: enforcement reads stale data

This one is **inferred from the mechanism, not directly observed** — worth confirming before acting on it.

In WAL mode a reader sees the last *committed* state. The outer connection's writes are uncommitted for the whole groups phase, so the enforcers' separate connections should be reading the database **as it was before this run's groups were processed**. `get_all_managed_channels(conn)` at `keywords.py:107` would then be enforcing against a pre-run snapshot — channels created or updated during this run would be invisible to it.

If that holds, enforcement has been operating on stale state whenever it *did* succeed, not just when it deadlocked.

## Suggested fix

Move enforcement (and the cleanup steps) **outside** the outer connection's `with` block, so the groups phase commits before post-processing begins.

This addresses both problems at once: the write lock is released, and the enforcers' connections see the committed results of the run they are meant to be enforcing. It also matches the existing design intent — these steps are described as post-processing, and they already open their own connections, so no signatures change.

The alternative — threading the outer `conn` down into the enforcers — keeps one transaction but requires touching every enforcement and cleanup entry point, and leaves the whole phase in a single long-held write transaction.

A third option, lowering `busy_timeout`, only makes the failure faster. It does not fix it.

## Provenance — not #606

- Earliest occurrence: **2026-08-25 05:35:40**. Earliest commit on `perf/generation-hot-paths`: **2026-08-25 11:40:19**, ~6 hours later.
- That branch never touches `consumers/enforcement/`, `database/channels/`, or `database/connection.py` (`git diff --stat dev...HEAD` over those paths is empty).
- Its only change to `processor.py` is the `aggregate_xmltv` flag, which *removes* a read.

## Reproduction

Configure an exception keyword with `behavior = 'ignore'` that matches a stream currently attached to a managed channel, then run a full generation. Enforcement will stall 30s and log `[KEYWORD_ERROR] database is locked`.

## Notes for whoever picks this up

- Worth adding a regression test that runs enforcement with the group processor's connection open and asserts the write lands — the current tests exercise enforcement standalone, where the outer connection does not exist, so the deadlock is invisible to them.
- While in here: `[RESOLVER] Pattern mode` (`lifecycle/dynamic_resolver.py:406`) and `[EVENT_EPG] Template resolution` (`event_group_processor/xmltv.py:141`) log ~1,500 near-duplicate INFO lines per run between them, which rotates the main log every ~30 minutes and leaves only ~2 hours of history. That is what made this failure hard to trace back. Demoting both to DEBUG is a one-liner and worth doing separately.

lpukatch:
Fixes #607. Bead: `teamarrv2-6nd2`.

Found while watching a live generation during the #606 soak. **Not caused by #606** — first occurrence was six hours before that branch's earliest commit, and it never touched these files.

## The bug

Every generation stalled exactly 30 seconds and then dropped the write:

```
13:32:16 | INFO  | [FILTER] Group 'Live | Ligue1 🇫🇷': 5 input → 0 passed
13:32:46 | ERROR | [KEYWORD_ERROR] database is locked
  keywords.py:139 remove_stream_from_channel → streams.py:102 conn.execute
```

30s is exactly `PRAGMA busy_timeout=30000`. Observed 11 times in one day, on every run — **31s of a 220s groups phase**.

It is a self-deadlock between two connections on one thread:

1. `process_all_groups` holds a connection for the whole groups phase and **writes** through it (`store_group_xmltv`, per group), so SQLite takes the WAL write lock at the first write.
2. There is no `conn.commit()` in `processor.py` — the commit is at context exit.
3. Enforcement ran *inside* that block, and every enforcer opens its **own** connection.
4. That connection's `UPDATE` blocked on a lock that could not be released until the block it was blocking exited.

No timeout resolves that. It failed 100% of the time whenever a step actually had a write to perform.

## The fix

Move enforcement after the `with` block, so the groups phase commits first.

`_run_enforcement` no longer takes a connection. It only ever used one for `subscription_cleanup`, which now opens its own; the other five steps already did. Keeping the connection out of the signature makes the correct call site the only convenient one — taking a live `conn` is what made it natural to call from inside the block in the first place.

## A second bug, filed as inferred and now confirmed

I flagged in #607 that the enforcers were probably also reading **stale** data, but hadn't proven it. I have now:

```
enforcer's view WHILE groups phase is mid-transaction : './data/teamarr.xml'
enforcer's view AFTER the groups phase commits        : 'written-by-groups-phase'
```

A second connection sees only committed state, so while the groups phase held its transaction open the enforcers were reading the database **as it looked before the run** — enforcing against channels that no longer matched reality. That was true even on runs where the write did not deadlock. The same reordering fixes it, and there is a test for it.

## Blast radius was wider than keywords

The same pattern sits in `enforcement/ordering.py:87` and four sites in `lifecycle/cleanup.py`. Only the step with actual work trips it, which is why it stayed hidden for months: in the run I captured, `cleanup_disabled_groups` ran immediately after the failure, found nothing to delete, and never attempted a write. All six steps are now outside the transaction.

The bug is likely months old and became *reachable* when a data condition changed — an `ignore`-behavior keyword started matching a live stream. The errors log goes back to 2026-04-03 with no prior occurrence.

## Ordering safety

Enforcement moves streams and deletes channels; it never rewrites a group's stored guide. So the XMLTV aggregation that now precedes it is unaffected, and `get_all_group_xmltv` still reads its own writes through the same connection as before.

## Tests

The existing enforcement tests run it standalone, with no outer connection, so this was invisible to them. Four new ones that are not:

| Test | Pins |
|---|---|
| `test_a_second_connection_cannot_write_under_an_open_write_txn` | the SQLite constraint itself, so the reasoning behind the ordering is verifiable rather than asserted in a comment |
| `test_enforcement_runs_after_the_groups_connection_is_closed` | an enforcer's own connection can write |
| `test_enforcement_sees_the_run_it_is_enforcing` | the stale-read half |
| `test_run_enforcement_does_not_borrow_a_caller_connection` | signature guard rail |

**I verified the middle two fail against the old ordering** (`2 failed, 1 passed` with the previous `processor.py` restored).

## Verification

`ruff` clean, pyright `0 errors`, **2335 tests pass** (+4), frontend builds.

## Not in this PR

`[RESOLVER] Pattern mode` and `[EVENT_EPG] Template resolution` log ~1,500 near-duplicate INFO lines per run between them, rotating the main log every ~30 minutes and leaving only ~2 hours of history — which is what made this hard to trace back to its first occurrence. Worth demoting to DEBUG separately; happy to fold it in if you'd rather.


lpukatch:
Folded in as `7d42e821` — but with a correction and a change of approach, because the premise I gave you was partly wrong.

## Correction: demoting does not fix the log retention

I said this would fix the ~30-minute rotation / ~2-hour history. It won't. `logging.py:154` pins the file handler at DEBUG unconditionally, with an explicit comment at line 36 — *"handlers are deliberately NOT adjustable: teamarr.log always captures DEBUG."*

Measured on the live pod, that file is **95% DEBUG** — 39,443 of 41,564 sampled lines. The lines in question are already in it and would only change label. Rotation is a DEBUG-volume problem and needs its own fix (bigger `maxBytes`, more backups, or making the file handler's level configurable). **Not addressed here**, and worth its own issue if you want the history back.

What it does help is the console / `kubectl logs` stream, which `LOG_LEVEL` actually governs: **2,975 → ~1,456 INFO lines per run, 51% fewer**.

## `[RESOLVER]` — deduped, not demoted

You asked me to demote only if genuinely not useful, and this one failed that test in an interesting way. The mapping is answered from `_groups_by_name`, so it was re-derived once per channel: 722 lines, 330 character-identical. But the **distinct** mappings are the diagnostic content — "which group did MLB land in?" is exactly what you'd scan for. Demoting would have thrown away the useful part along with the noise.

So: first occurrence of each distinct outcome stays at INFO, repeats go to DEBUG. **722 → 15 lines, nothing lost.** A changed `group_id` for the same name counts as a new outcome and is announced again.

`DynamicResolver` is run-scoped (`service.py:208`), so the dedupe set has the right lifetime.

## `[EVENT_EPG]` template resolution — demoted

These two genuinely are debugging detail. Per-event and per-group, for answering "why did THIS event get THAT template" — which is a debugging question, and DEBUG is where you go to answer it. Their own comments already say *"Log template resolution context"*. 812 lines (750 per-event + 62 identical copies of the subscription summary).

## Tests

Three new ones pinning the dedupe, aimed at the failure mode that would make this a bad trade — silently dropping a distinct mapping: every distinct mapping is announced, 330 repeats produce exactly 1 INFO (and 329 recoverable DEBUG), and a changed `group_id` re-announces.

`ruff` clean, pyright `0 errors`, **2338 tests pass** (+3), frontend builds.

PR #612, opened 2026-08-25
perf: build country resolver and identity index once per run (#609)

### LINKED ISSUE #609: Hoist per-group matcher setup to per-run (~19s of the groups phase)
Measured against a live install during the #607 soak. **Filed rather than implemented** — the win is real but it carries three specific hazards that need handling first, written up below so whoever takes it does not have to rediscover them.

## The cost

That install runs **343 event groups**. Each one constructs a fresh `StreamMatcher`, which constructs a fresh `TeamMatcher`. Measured against its own database (11,095 `team_cache` rows, 295 known leagues):

| Per group | × 343 groups |
|---|---|
| `StreamMatcher`/`TeamMatcher` construction — 21 ms | **7.3s** |
| `TeamIdentityIndex` rebuilt from DB — 19 ms (11,079 rows → 17,529 entries) | **≤6.4s** |
| Warm prefetch plan pass — 16 ms | **5.6s** |

~19s of a 200s groups phase. The identity index also re-runs a `SELECT DISTINCT` over `team_cache` (22.7 ms) each time it is built, on top of the normalisation cost.

## Why it looks easy

`TeamMatcher.__init__(service, cache, db_factory, days_ahead)` takes **no group-specific arguments at all**. Group identity arrives per-call through `MatchContext(group_id=...)`, never through construction. It is rebuilt 343 times only because `StreamMatcher` — which *is* group-specific (`group_id`, `custom_regex_*`, `include_leagues`, `epg_index`) — happens to build it.

So the shape of the fix is: let a run build one `TeamMatcher` (or at least one `TeamIdentityIndex` + alias cache) and pass it into each group's `StreamMatcher`.

## Hazard 1 — identity-keyed memo (do this first)

`TeamMatcher._candidates_memo` invalidates on **object identity**:

```python
if self._candidates_source is not prefetched_events:
    self._candidates_source = prefetched_events
    self._candidates_memo = {}
```

Today this is airtight, because a per-group `TeamMatcher` only ever sees one prefetch dict, alive for that group's whole lifetime.

Shared across groups it stops being airtight. Group A's dict is freed when A finishes; group B's new dict can be allocated **at the same address**; CPython reuses freed memory; `is not` returns False; the memo never resets; **B silently matches against A's candidate events**. Low probability, wrong answers, no error, no log line.

Key the memo on an explicit monotonic token (or the run's generation counter) instead of object identity before sharing anything.

## Hazard 2 — an accidental bound becomes unbounded

`TeamMatcher._alias_resolve_cache` is an unbounded `dict` keyed `(team_name, league)`. Per-group construction is currently the only thing capping it. Shared across 343 groups it grows for the entire run. Needs an explicit bound (or an `lru_cache`).

## Hazard 3 — latent thread-safety

The group loop is sequential today, so shared mutable memos are safe. They stop being safe the moment anyone parallelises groups — which is plausible given the direction of #605. Whatever is shared needs an explicit comment stating the sequential-groups assumption it relies on, so a future change to the loop does not silently break matching.

## Suggested order

1. Replace the identity-keyed memo invalidation with an explicit token. **This is worth doing on its own merits** — it is a correctness trap sitting in the code waiting for exactly this refactor.
2. Bound `_alias_resolve_cache`.
3. Then hoist: build the identity index and alias cache once per run, pass them into each group's `StreamMatcher`.

## Related

- The prefetch plan pass (third row above) allocates **295 leagues × 34 dates = 10,030 `_PrefetchSlot` objects per group**, even when every one of them is reused from `shared_events`. That wants an early-out, and is largely independent of the hoisting work.
- Introduced alongside #605 / #606.

lpukatch:
Closes #609. Bead: `teamarrv2-gh6v`.

## What

A `TeamMatcher` is constructed per event group — **343 groups** on the install this was measured against. Two of the things it built do not vary by group. Measured against that live instance:

| | per group | × 343 |
|---|---|---|
| `CountryNameResolver()` | 39.2 ms | **13.4s** |
| `TeamIdentityIndex.from_db()` | 61.7 ms | **21.2s** |
| | | **34.6s → 0.101s** |

(Idle-machine figures were 17.4ms / 46.9ms; the above are under real load, which is the honest number.)

**Caveat on the split:** the resolver is built in every `TeamMatcher` constructor, so its ~13.4s is a hard floor. The identity index is lazy — only built when a group actually reaches the fixture gate — so its ~21.2s is an upper bound, not a guarantee.

## Why these two are safe to share

`CountryNameResolver` is genuinely read-only: every write to `_map` happens inside `_build()` during construction, and `resolve()` only reads. Its own docstring already called the mapping *"static for the lifetime of the process"* — it just wasn't treated that way. There is a test asserting `resolve()` does not mutate it.

`TeamIdentityIndex` is an unfiltered scan of `team_cache` (11k rows live) plus normalising every surface form. Now memoized per process behind a 900s TTL — the same shape as `_cached_team_identity` in `services/sports_data.py`, for the same reason: team identity is effectively static, and the TTL bounds how long a mid-run cache refresh stays invisible.

## What I deliberately did NOT do

**The `TeamMatcher` itself is still per-group.** That was the design in the issue, and it turned out to be the wrong shape — sharing the matcher would have dragged its mutable memos (`_alias_resolve_cache`, `_candidates_memo`) across group boundaries, which is exactly hazards 2 and 3. Sharing only the two read-only objects gets essentially the whole win and makes both hazards inapplicable rather than mitigated.

(Hazard 1, the identity-keyed memo, was withdrawn earlier — verified sound. `_candidates_source` holds a strong reference, so the address-reuse scenario I described cannot occur.)

## Two things the sharing did require

- **`TeamIdentityIndex.resolve`'s memo is now bounded.** It outlives a single group, so it no longer gets a fresh start each group. Cleared wholesale at the cap, matching `_TEAM_IDENTITY_MEMO`'s existing idiom.
- **An autouse conftest fixture resets the shared index between tests.** Every test gets its own temp database, so without it a test could be served an index built from a *previous test's* `team_cache` — silently matching against the wrong teams. This is the kind of cross-test leak that produces confusing failures much later, so it is worth having even though nothing currently trips it.

The empty-`team_cache` path still returns without caching, so a run starting before the first cache refresh is not stuck with "no index" for a whole TTL window.

## Tests

14 new: sharing, read-only-ness, TTL and reset, single-flight under concurrent first use, the memo bound, that the per-group mutable memos are still *not* shared, and that resolution results are identical to a standalone index.

One correction they forced: a fresh temp DB is **not** empty — `schema.sql` seeds 22 `team_cache` rows. My "empty install" tests were wrong-premised and now clear the table explicitly rather than assuming.

## Verification

`ruff` clean, pyright `0 errors`, **2361 tests pass** (+14), frontend builds.


egyptiangio:
## Review — second pass

Verified: ruff clean, 2361 tests pass, resolver-build cost reproduced at 8–50ms/instance on this box (claimed magnitudes plausible, the ~35s win is real). The safety argument holds up: `CountryNameResolver.resolve` is a pure `_map.get` with all writes confined to `_build()` (and the immutability test pins it); the index memo's double-checked locking is right; failure/empty results are correctly not cached so a mid-run first seed becomes visible; per-group mutable memos stay per-matcher; the conftest reset keeps the process-wide memo out of cross-test pollution. Good tests — single-flighting and the empty-not-cached case are exactly the right things to pin.

**One change requested: `reset_identity_index_cache()` has no production caller.** Its docstring promises it to "callers that have just reseeded team_cache", but only tests use it — so a cache refresh leaves the fixture gate on a stale index for up to 15 minutes. That regresses the refresh-then-generate sequence: before this PR every group rebuilt the index live, so a refresh was always visible immediately. And unlike `_TEAM_IDENTITY_MEMO` (TTL-only, but display-enrichment only), this index **vetoes** — a stale league membership yields `FIXTURE_NOT_IN_LEAGUE`, a silent missing match, the #599/#601 failure class the goax epic worked to keep at zero false vetoes. Fix is one line: call the reset at the end of team-cache seeding/refresh (both the full and per-league paths), plus a test that a refresh invalidates the shared index. Consider clearing `_TEAM_IDENTITY_MEMO` there too for consistency, though it's cosmetic by comparison.

Minor, no action needed: the memo ignores `db_factory` identity (fine — one DB in production, tests handled via conftest), and a fresh install with an *erroring* team_cache read retries per group with a warning each time (same as before this PR).

**Path to merge:** wire the reset into cache refresh + one test, then good to merge.

lpukatch:
Fixed in `5c44d2aa`. You were right, and it was a regression I introduced rather than a gap in scope — `reset_identity_index_cache()` had no production caller. I wrote it for tests and then documented it as though production used it, which is how the false promise got past me.

The severity framing is what makes this a fix rather than a polish item, and I've written it into the code: this index **vetoes**. A stale league membership yields `FIXTURE_NOT_IN_LEAGUE` — a silently missing match, the failure class goax worked to hold at zero false vetoes. That is genuinely unlike `_TEAM_IDENTITY_MEMO`, which is TTL-only but only backfills display fields.

## What changed

Added `database.team_cache.invalidate_team_identity_caches()` as a single entry point — so a caller cannot half-invalidate — and wired it into every path that writes the table:

| Path | |
|---|---|
| `CacheRefresher._save_league_teams` | per-league refresh |
| `CacheRefresher._save_cache` | full refresh |
| `purge_league_cache_rows` | custom-league delete, guarded on actually having removed team rows |

I went slightly past the two you named: `purge_league_cache_rows` deletes `team_cache` rows too, so it leaves the gate vetoing against rows that no longer exist. Same class of bug, and there is a test for it.

It clears `_TEAM_IDENTITY_MEMO` as well, as you suggested. Cosmetic next to the veto, but it reads the same rows and dropping them together is one less thing for a future caller to get half-right.

**Not wired: `database/seed.py`.** It runs inside `init_db`, before anything can have built an index, so there is nothing to invalidate. Flagging so the omission reads as deliberate.

Lives in `database/team_cache.py` — the module that owns the table — with function-local imports, since both consumers import from it and module-level would be circular.

## Tests

Seven, and **I verified two of them fail with the invalidation removed** — i.e. against exactly the state this PR shipped in. Beyond the refresh-invalidates case you asked for, they also pin that an empty refresh does not leave the gate permanently disabled, and that invalidation has not defeated the memoization it exists to guard (that one would have caught an over-eager fix).

`reset_identity_index_cache()`'s docstring now points at the helper rather than promising a contract it did not have.

`ruff` clean, pyright `0 errors`, **2368 tests pass** (+7), frontend builds.

Issue #91, opened 2026-01-18
[Enhancement]: NCAA Team Setup/Selection


Sp0rtsFr3ak:
### Teamarr Version

V2 (alpha)

### Current behavior

Select all teams (600+), or select them individually.

### Proposed improvement

Keep the current functionality, but add filters to allow users to add teams by conference(s). 

### Why?

When setting up teams, allow users to filter/select by conference. Adding all 600-700 teams is likely unnecessary for most users, so adding/editing by conference would be helpful. If I had to guess, most users would probably select 100 to 150, that's a lot of work to setup when selecting individually. Most people will add by conference though.

egyptiangio:
Currently researching NCAA data sources (related to #84) including the possibility of building a web scraper for ncaa.com. We have a plan ready for implementing conference hierarchy using the ESPN provider's standings endpoint. Should be able to move forward on this soon once we finalize the data source approach.

egyptiangio:
Quick correction/update on this one. My earlier comment tied this request to the NCAA scraper research (#84) — that was wrong. Selecting teams by conference doesn't need a new data source: ESPN already exposes conference/standings groupings for the NCAA leagues we support, so this is a self-contained UX improvement to the team picker.

I've separated it from the (now closed) scraper work and it stays on the roadmap as its own item: surface ESPN's conference grouping in NCAA team setup so you can add a whole conference at once instead of scrolling 600+ teams individually. No ETA yet, but it's correctly scoped and decoupled now. Thanks for the clear write-up — the 100-150-teams-by-conference use case is exactly right.

egyptiangio:
API interrogation complete (2026-07-17) — conference selection is feasible for basketball and football, unsupported by ESPN for the rest.

**NCAAM / NCAAW — trivial.** One call: `site.api.espn.com/apis/site/v2/sports/basketball/{mens,womens}-college-basketball/groups` returns the full Division tree with **25 conferences and all 290 teams inline** (team id, displayName, abbreviation, logos). Conferences carry name + abbreviation (no numeric id) — abbreviation is the stable key.

**College football — different plumbing, verified working.** The site `/groups` endpoint only returns division buckets truncated to 25 teams, and the site `?groups=` filter is unreliable. The core API season tree is authoritative:
- `sports.core.api.espn.com/v2/sports/football/leagues/college-football/seasons/{yr}/types/2/groups/90/children` → FBS (80) / FCS (81)
[...]

PR #635, opened 2026-08-28
feat(providers): add optional Bullpen proxy support

### LINKED ISSUE #548: Optional bullpen proxy support for the provider layer
## Summary

Add opt-in support for routing provider JSON API calls through [bullpen](https://bullpen.direct), a caching proxy that fronts several of teamarr's upstreams (ESPN, TheSportsDB, NASCAR, MLB Stats, HockeyTech, Squiggle) behind an `X-Bullpen-Key` header.

## Scope

- Global `bullpen_enabled` + `bullpen_api_key` + `bullpen_base_url` (default `https://bullpen.direct`) settings, default off.
- Per-provider toggle (espn, squiggle, nascar, mlbstats, hockeytech, tsdb), default off. Supabase has no bullpen target and is untouched.
- JSON API routing only in this pass — no image/binary proxying (espn-cdn, flagcdn, mlb-static, ncaa targets are out of scope; can follow up later if needed).
- TheSportsDB: enabling its bullpen toggle also makes `is_premium` report `True` (bullpen is being treated as equivalent to holding a TSDB premium key), and TSDB requests route through bullpen's `thesportsdb` target.
- No fallback to the direct origin on bullpen failure (502/429/etc.) — errors surface like any other failed provider call.
- Settings page lives at `/bullpen`, deliberately with no sidebar/nav entry — reachable only by navigating to the URL directly.

## Out of scope (this pass)

- Image/binary proxying via bullpen's `/v1/img` target.
- Supabase provider (no bullpen target exists for it).

sethwv:
## Summary

Adds opt-in Bullpen proxy routing for supported provider API requests.

- Adds Bullpen settings and the `/bullpen` configuration page.
- Routes ESPN, Squiggle, NASCAR, MLB Stats, HockeyTech, and TheSportsDB through Bullpen when enabled per provider.
- Preserves direct provider access by default.
- Treats Bullpen-enabled TheSportsDB access as premium-capable.
- Documents configuration and provider behavior.

## Validation

- Full backend test suite
- Ruff
- Frontend production build

Closes #548


egyptiangio:
Clean wiring — factory-boundary config, registry-driven settings, default-off at all three levels. A few things before it leaves draft, most of which protect your keys and your quota:

**Must-fix**
1. **Key leak:** `ESPN_CORE_URL` (`espn/client.py:328`, the #91 season-tree calls) is `http://` and isn't rewritten, but `X-Bullpen-Key` is a default header on the shared client — so conference fetches send the bullpen key to ESPN in cleartext. Route via `espn-core` like the UFC URL; ideally only attach the header when the host is bullpen (see "future scale" below — this is the change that makes every later target safe by construction).
2. **Restart semantics:** `PUT /settings/bullpen` never calls `reinitialize_provider()`, so toggles (including *disable*) do nothing until restart. `display.py:157` does this for the TSDB key — same pattern, all six providers.
3. **Tests:** none ship; please cover `bullpen_rewrite` (incl. trailing-slash bases — HockeyTech `/feed/`, Squiggle `/`), config gating, header-by-host, `is_premium`, NASCAR; and add bullpen to `pairs` in `tests/test_settings_registry.py`.
4. **Drop `.beads/*.jsonl`** — co-dev exports don't round-trip into the Dolt DB (we lost a bead this way today).
5. **CodeQL:** the two `tsdb/client.py` alerts predate you (July) but landed in changed lines — log `endpoint` not `url` and it goes green.

**Should-fix**
6. Factory-bypassing constructions never see bullpen: `cache/refresh.py:246,599,724`, `custom_leagues.py:533` — the full cache refresh stays direct-to-origin.
7. `is_premium=True` also suppresses fallback providers (`league_mappings.py:479`) and enables the premium prewarm (`refresh.py:543`) — is bullpen's TSDB quota per-user or shared? Also: does bullpen substitute its own TSDB key? If so we should send `3`, not the user's.
8. Enable-without-key saves silently as a no-op; no way to clear a stored key (use the `_NOT_PROVIDED` sentinel); validate `base_url` as `https://`.
9. Prefer `get_bullpen_settings(conn)` over raw SQL in the factories; narrow the bare `except` (a wrong answer here changes where traffic goes).

**Docs**
10. One line in `bullpen.md` that it's your service and what it sees when enabled; `database.md:33` still says 123 columns; README/guide page per the docs table. And your call with @egyptiangio on whether `/bullpen` gets a nav entry.

**Future scale — what would make adding targets cheap**

The per-provider shape (`bullpen_<provider>_enabled` + `bullpen_rewrite(base, target, cfg)`) is the right one, and I looked at what's left:

- **Bell Media (CFL)** is the same 4-line change as the others — `BaseHTTPClient`, one static base, no key, plain JSON. Happy to see it in this PR or a follow-up if bullpen grows a `bellmedia` target.
- **Supabase (CBL)** should stay out, as #548 says: it scrapes the league site for a per-league host + key, so there's no single upstream to front.
- **Images/logos** are a different job — Teamarr mostly hands those URLs to Dispatcharr/clients rather than fetching them, so proxying means rewriting stored logo URLs, which would put the bullpen key into every EPG/M3U consumer unless bullpen serves images key-less.

Two structural asks so the next target is a config entry rather than a review:

- **Attach `X-Bullpen-Key` per-request, keyed on host**, not as a client-wide default header. That's what turns #1 from "we caught it this time" into "can't happen" — any future un-rewritten URL simply goes direct without the key.
- **One place that knows the target table** (provider → bullpen target name + origin base), so a new provider is a row there plus its `_enabled` column, and the "which paths are proxied" doc can be generated from it rather than hand-maintained.

sethwv:
Addressed the must-fix review items in `31bfc633`:

- Bullpen credentials are now attached per request only when the destination host is the configured proxy. Direct-origin requests cannot receive `X-Bullpen-Key`.
- ESPN season-tree calls now route through `espn-core`; NASCAR uses the same host-scoped header check.
- Bullpen settings updates reinitialize all proxied providers immediately.
- Added coverage for rewrite normalization, config gates, proxy-only headers, anonymous mode, ESPN core, NASCAR, TSDB premium behavior, settings parity, and provider reinitialization.
- Bullpen can now operate anonymously when enabled by its administrator: the key is optional, no header is sent without one, and an existing key can be cleared in the UI.
- Added a Bullpen privacy disclosure to the provider reference.

Also rewrote the two prior commits to remove their JSONL changes while retaining the tracked JSONL files in the repository.

Validation: `pytest` (2450 passed), `ruff check teamarr tests`, and `npm run build`.

Issue #653, opened 2026-08-29
Apostrophes split team names — normalize_for_matching and normalize_text disagree, and both run


lpukatch:
Discovered auditing NCAAF matching on 2026-08-29 (v2.13.2).

## Impact

Any team name containing an apostrophe fails to match, even with a perfectly parsed stream. `Hawai'i vs. Stanford` scored 100.0 on `_score_teams_against_event` directly, yet returned `no_event_found` through the real pipeline.

## Mechanism

Two normalizers disagree on apostrophes, and the matcher applies both in sequence:

```python
normalize_for_matching("Hawai'i")  -> 'hawai i'   # normalizer.py:  [^\w\s] -> " "
normalize_text("Hawai'i")           -> 'hawaii'    # fuzzy_match.py: apostrophe deleted
```

`_match_against_events` (`team_matcher.py:1123`) runs `normalize_for_matching` first, then `_score_teams_against_event` runs `normalize_text` on the already-split result. The split is unrecoverable:

```
'hawai i' vs 'hawaii rainbow warriors' = 46.7   <- what the matcher actually compares
'hawaii'  vs 'hawaii rainbow warriors' = 100.0  <- single normalizer
```

46.7 is below `BOTH_TEAMS_THRESHOLD` (60), so the whole event is rejected.

## Confirmed affected

```
Hawai'i         -> 'hawai i'
American Int'l   -> 'american int l'   (72.7 vs "American International Yellow Jackets")
G'town Col      -> 'g town col'
```

Real failure from `epg_failed_matches`:
`NCAAF 22: Hawai'i vs. Stanford @ Aug 29 6:00PM ET  t1="Hawai'i" t2='Stanford'  no_event_found`

## Fix

Make `normalize_for_matching` delete apostrophes (`'` and `’`) before the punctuation-to-space rule, matching `normalize_text`. Worth auditing whether the two normalizers should differ at all.

lpukatch:
Bead: `teamarrv2-ibh5`

lpukatch:
PR #659 opened against `dev` — CI green, left open for review. Paired with #652 since both live in the same shared normalization path.

`normalize_for_matching` now deletes apostrophes and backticks before the punctuation rule, matching `normalize_text`. Hawai'i goes 46.7 → 100.0 against "Hawai'i Rainbow Warriors".

egyptiangio:
Pre-implementation review. Diagnosis confirmed; the fix as worded misses one form.

[...]

Issue #652, opened 2026-08-29
Pipe-delimited metadata leaks into team2 — neither classifier nor matcher resolves it


lpukatch:
Discovered auditing NCAAF matching on 2026-08-29 (v2.13.2).

## Impact

All 22 failures in source `Sports | NCAAF Events` have pipe metadata in `team2`. Trimming `team2` at the first pipe takes the sampled streams from **1/4 to 4/4**.

```
t2='WAGNER | 8.29  | NEC FRONT ROW'            vs "Wagner Seahawks"  = 57.1   (BOTH_TEAMS_THRESHOLD = 60)
t2='TCU (IN DUBLIN, IRELAND) | 8.29  | ESPN'   vs "TCU Horned Frogs" = 38.3
t2='CHICAGO STATE | 8.29  | TV TBA'            vs "Chicago State Cougars" = 78.8
t2='UNLV | 8.29  | FOX'                        vs "UNLV Rebels"      = 53.3
```

Clean names score 100.0. The ones that match today do so by luck — UNLV survives only via the abbreviation path, Virginia only because 61.5 clears 60 by 1.5.

## Mechanism — each side thinks the other handles it

`classifier.py:1026` deliberately preserves pipe content:

> Handle "|" separator - preserve pipe content for fuzzy matching disambiguation.
> The matcher will try both sides of the pipe and pick the one that matches.

`team_matcher.py:1827` asserts the opposite:

> Pipe-separated content (e.g., "Sacramento Kings | Golden 1 Center") is handled
> naturally by token_set_ratio which finds best token overlap.
> No explicit pipe resolution needed.

There is no pipe handling anywhere in `team_matcher.py` (grep confirms one comment, no code). The second claim is false: `token_set_ratio` penalises the extra tokens on the stream side.

There *is* a parenthetical fallback (`_prepare_stripped_fallback`); a pipe-segment fallback is the missing sibling.

## Secondary: dot-separated dates are not recognised

`DATE_PATTERNS` (`normalizer.py:260`) supports `/` and `-` only. `8.29` is never extracted, so it is both lost as a date **and** left inside the team name:

```
"... AT WAGNER | 8.29 12:00PM | NEC"  -> t2='WAGNER | 8.29  | NEC FRONT ROW'   date=None
"... AT WAGNER | 8/29 12:00PM | NEC"  -> t2='WAGNER |   | NEC FRONT ROW'        date=2026-08-29
```

## Fix

[...]

Issue #651, opened 2026-08-29
[1080p] resolution tag is not stripped from team names — entire source matched 0/30


lpukatch:
Discovered auditing NCAAF matching on 2026-08-29 (v2.13.2).

## Impact

Source `USA: NCAAF PACKAGE [1080p]` matched **0 of 30** streams. The same package without the tag (`US| NCAAF PACKAGE`) matched 21/30.

Measured against the real ESPN slate for 2026-08-29:

```
group 164 today-slate streams, AS-IS:    0/21
group 164 with [1080p] stripped:        21/21
```

## Mechanism

`_clean_team_name` (`teamarr/consumers/matching/classifier.py`) strips:

- `HD|SD|FHD|4K|UHD` quality indicators
- bracketed **pure digits**: `\s*[\(\[]\d+[\)\]]\s*$` — matches `[2]`, not `[1080p]`

`[1080p]` falls between the two rules and survives into `team2`:

```
t2='Wagner  [1080p]'      token_set_ratio vs "Wagner Seahawks"       = 66.7   (BOTH_TEAMS_THRESHOLD = 60 ... but min() across both sides fails)
t2='Notre Dame [1080p]'   token_set_ratio vs "Notre Dame Fighting Irish" = 76.9
```

Sample of affected rows from `epg_failed_matches`:

```
NCAAF 16: Robert Morris at Wagner 12pm [1080p]     t2='Wagner  [1080p]'
NCAAF 29: The Citadel at Wofford 6pm [1080p]       t2='Wofford  [1080p]'
NCAAF 40: Montana State at Utah Tech 10pm [1080p]  t2='Utah Tech  [1080p]'
```

## Fix

Extend the quality-indicator strip to resolution tags (`1080p`, `1080i`, `720p`, `2160p`, `480p`, …), bracketed or bare, at either end. Keep it anchored so a legitimate digit-bearing team name (49ers, 76ers) is untouched.

lpukatch:
Bead: `teamarrv2-eg8j`

lpukatch:
Parking this — closing in favour of a per-source regex.

## Why the app-side case is weaker than I first argued

I claimed 33,572 affected failures. That was **presence of the tag, not causation**, and it badly overstated the impact. This install has natural twins — the same content sourced with and without the tag — and they mostly perform identically:

```
TAGGED                            rate      UNTAGGED twin                rate
[...]

Issue #650, opened 2026-08-29
Fixture gate falsely vetoes 27% of college football (NCAA soccer bare-school names shadow every other sport)


lpukatch:
Discovered auditing NCAAF matching on 2026-08-29 (v2.13.2).

## Impact

512 of 1026 failure records in a support bundle are `fixture_not_in_league` (49 distinct streams). Running the real `TeamIdentityIndex` against a live `team_cache` over the full ESPN slate for 2026-08-29:

```
today CFB fixtures: 73    gate OK: 39    DEFER: 14    VETOES college-football: 20
  Fairmont State vs Dayton                 -> ['usa.ncaa.w.1']
  San José State vs USC                    -> ['usa.ncaa.w.1']
  New Mexico State vs Florida State        -> ['usa.ncaa.w.1']
  Jacksonville State vs North Dakota State -> ['usa.ncaa.w.1']
  Campbell vs East Tennessee State         -> ['usa.ncaa.m.1', 'usa.ncaa.w.1']
  Montana State vs Utah Tech               -> ['college-softball']
```

27% of the slate is unmatchable by any stream naming the teams by bare school name, however clean the stream name is. FBS games included.

## Mechanism

`usa.ncaa.w.1` (NCAA Women's Soccer, 293 rows) and `usa.ncaa.m.1` (Men's, 200 rows) store the **bare school name as the full `team_name`** — "Fairmont State", "Eastern Washington", "VMI" — because ESPN publishes those leagues without mascots.

Per #619, only a full name is an exact identity. So `"Fairmont State"` resolves **exactly** to the soccer row and **narrows** the identity set to `{usa.ncaa.w.1}`. The football row is `"Fairmont State Falcons"`, where `"Fairmont State"` is a city/short-name prefix — a *partial* reading that widens but never narrows. One narrowly-resolving side is enough to veto the whole fixture.

`knows_league("college-football")` is True (760 cached CFB teams), so that safety valve does not fire.

The #619 rule assumed bare-school naming was a `team_short_name` phenomenon. In these two leagues it is the canonical `team_name`, so those rows act as exact identities that shadow every other sport's "School Mascot" rows.

## Scope

Not football-specific. Any college sport matched by bare school name (NCAAB, baseball, softball, volleyball) is exposed to the same shadowing.

## Verification

```python
from teamarr.consumers.matching.identity import TeamIdentityIndex
[...]

PR #655, opened 2026-08-29
fix(matching): stop mascotless NCAA soccer rows shadowing college football in the fixture gate (#650)

lpukatch:
Closes the matching half of #650.

## Problem

`usa.ncaa.w.1` / `usa.ncaa.m.1` (NCAA soccer, ~490 cached rows) publish **no mascots**, so their full `team_name` is the bare school — `"Fairmont State"`, `"Eastern Washington"`, `"VMI"`. Per #619 a full name is an *exact* identity, so those rows claim the school-only form outright.

The football row is `"Fairmont State Falcons"`. It should have been reachable as a *partial* reading, but it was not reachable at all:

| | value |
|---|---|
| `team_name` | `Fairmont State Falcons` |
| `team_short_name` | `Fairmont St` |

The existing prefix rule only derives a prefix when the short name is a **suffix** of the full name (`Milwaukee Brewers` / `Brewers` → `milwaukee`). ESPN abbreviates the **school** for college, never the mascot, so `Fairmont St` is not a suffix of `Fairmont State Falcons` and the rule never fires. `_partial["fairmont state"]` stayed empty.

Result: one narrowly-resolving side collapsed the fixture to `{usa.ncaa.w.1}` and vetoed `college-football`.

```
idx.resolve("Fairmont State")                    # exact=True, leagues={usa.ncaa.w.1}
idx.fixture_leagues("Fairmont State", "Dayton")  # {usa.ncaa.w.1}  -> vetoes CFB
```

`knows_league("college-football")` is True (760 cached CFB teams), so that safety valve could not fire.

## Fix

Register every **≥2-token leading run** of a full name as a partial reading. Every prefix rather than just the mascot-dropped one, because mascots run to two words as often as one (`Central Connecticut Blue Devils`).

Safe by construction: partial readings only **widen** an identity set, and the gate is veto-only — so this can withdraw a false veto but never manufacture a match. Prefixes stop at two tokens so a bare `north`/`saint` never enters thousands of teams.

## Measured

Against the **live** `team_cache` over the complete ESPN slate for 2026-08-29 (73 fixtures, groups 90 + 35):

| | before | after |
|---|---|---|
| gate OK | 39 | **64** |
| defer | 14 | 8 |
| **false vetoes** | **20** | **1** |

The one remaining is an ESPN data defect, not a gate defect — their CFB row is literally `Oklahoma Panhandle OK PANHANDLE ST` (no "State"), so no prefix of it spells the school.

For scale: 512 of 1026 failure records in the reporter’s support bundle were `fixture_not_in_league`.

## Tests

`TestMascotlessLeaguesDoNotShadowMascotedOnes` in `tests/matching/test_fixture_gate.py` — 7 cases covering the school-only form, two-word mascots, both directions of the fixture, the two-token bound, and that the original Lightning/Rays crosstalk veto still fires. The existing corpus (0 false vetoes / 200, 322/322 crosstalk rejected) did not cover FCS/D-II football or bare-school naming.

Gates: `2557 passed`, `ruff` clean, frontend builds.

## Not in this PR

Three other bugs found in the same audit, filed separately: #651 (`[1080p]` tag, one source scored 0/30), #652 (pipe metadata in team2), #653 (apostrophes split team names).


egyptiangio (approved):
Reviewed the mechanism in `identity.py`: new keys land only in `_partial`, which `_resolve_uncached` only ever unions onto exact hits or returns as non-exact — so this can withdraw a veto but never create one. Two-token floor means mascotless 2-token rows register nothing. Ran `tests/matching` on the branch: 494 passed, corpus unchanged.

One cosmetic nit, no action needed: when the short-name rule and this rule derive the same prefix ("South Dakota Coyotes"/"South Dakota") the identity is appended twice; `dict.fromkeys` dedupes on read.

Issue #645, opened 2026-08-29
Deprecate per-group event_group sub-runs in processing_runs


egyptiangio:
Every full generation writes one `full_epg` row plus one `event_group` sub-run row per source in `processing_runs` (33/run on a 32-source install; 23k sub-run rows vs 700 full runs). Nothing ever runs a single group on its own — `process_event_group` / `GroupService.process_group` / `GroupService.process_all_groups` have zero callers.

The sub-run rows are load-bearing only by accident:
- `epg_matched_streams` / `epg_failed_matches` are keyed on sub-run ids; a full run finds its details by a **timestamp-window** join (`stats.py`), not a key.
- `streams_cached` lives only on sub-runs (#312), so `get_current_stats` and the lifetime fold each carry a second query for it.
- `get_match_stats_summary(run_id=<full run>)` queries details by the full-run id → `by_group` / `by_league` are always empty.
- "latest run" (`run_id` omitted) resolves to the last *sub-run*, not the full run.
- `cleanup_old_runs` never prunes the detail tables → unbounded growth.

Change: one run, one row. Groups report into the parent run (per-group breakdown in `extra_metrics.groups`), details keyed on the full-run id, #312 special-casing and timestamp joins removed, detail rows pruned with their run, dead entry points removed. One migration re-keys existing detail rows and folds legacy cache hits.

egyptiangio:
Bead: `teamarrv2-xesm`

egyptiangio:
Landed on dev: f4236956 (PR #646)

egyptiangio:
Released in [v2.14.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.14.0).

PR #658, opened 2026-08-29
feat(support-bundle): surface matching outcomes as signals (#657)

### LINKED ISSUE #657: Support bundle reports "No automatic signals were found" on an install where a source matched 0/30 and 82% of streams failed
From the 2026-08-29 NCAAF matching audit (see #650–#653). The bundle that surfaced four real matching bugs led with **"No automatic signals were found."**

## What the bundle said vs. what was in it

`AGENTS.md` → *Signal Summary: No automatic signals were found.*

Derived from `recent-runs.json` in that same bundle (latest run per group):

```
group 9      matched=10   of 35   ( 29%)
group 10     matched=4    of 7    ( 57%)
group 11     matched=0    of 30   (  0%)   <-- dead source
group 12     matched=7    of 27   ( 26%)
group None   matched=0    of 18   (  0%)   <-- dead

install-wide: matched=21 unmatched=96  (18% matched)
```

And from `recent-run-stream-details.json`:

```
total failed stream records: 1026
  fixture_not_in_league   512
  no_event_found          510
  team2_not_found           4
```

Two sources at literally zero. 82% of streams unmatched install-wide. One failure reason accounting for half of all failures. No signal.

## Cause

All six rules in `SupportBundleService._signals()` are about **configuration or infrastructure**:

`dispatcharr_disabled` · `no_enabled_sources` · `teams_without_template` · `latest_run_not_completed` · `media_server_refresh_failing` · `no_managed_channels`

None looks at **matching outcomes**. `_signals()` is called as:

```python
"signals": self._signals(settings, sources, teams, channels, runs)
```

`runs` already carries `streams.matched` / `streams.unmatched` per run, so a match-rate signal needs **no new collection** — it is pure derivation over data the bundle already ships. The per-reason breakdown lives in the separately-collected stream details, which `_signals()` is not passed.

This matters more than a missing nicety: "no signals found" is the first line a triager reads, and here it actively pointed away from a totally dead source.

## Proposed

Two new codes:

- `source_matching_zero` (**critical**) — a source whose latest run fetched streams and matched none. Evidence: group id/name, stream count.
- `source_matching_degraded` (**warning**) — latest-run match rate below a floor with a meaningful denominator, so a small/quiet source does not cry wolf.

Optionally `match_failure_reason_concentrated` (info) if the failed details get threaded into `_signals()` — worth it, since the dominant reason code is what actually named the bug in this audit.

Thresholds should be picked against real bundles, not guessed; off-season sources legitimately sit at zero matched with zero streams, and the zero-stream case must stay silent.

## Contract

Per the Support Bundle Contract in `CLAUDE.md`, adding signal codes requires updating implementation, `tests/test_support_bundle.py`, `docs/guide/troubleshooting.md` (line 135 enumerates the codes), the bundled `AGENTS.md`, and the CLAUDE.md instruction — same change.

lpukatch:
Closes #657.

## Problem

All six existing rules in `_signals()` are about **configuration or infrastructure** — `dispatcharr_disabled`, `no_enabled_sources`, `teams_without_template`, `latest_run_not_completed`, `media_server_refresh_failing`, `no_managed_channels`. None looks at **matching outcomes**, which is the thing the app exists to do.

So the bundle that surfaced #650–#653 led with:

> **Signal Summary** — No automatic signals were found.

while carrying, in its own `recent-runs.json`:

```
group 11  matched=0 of 30   <-- dead source
group 12  matched=7 of 27
group  9  matched=10 of 35
install-wide: 21 matched / 96 unmatched  (18%)
```

"No signals found" is the first line a triager reads, and here it pointed away from a completely dead source.

## Change

`source_matching_health()` in `database/stats.py` (sibling of `media_server_health`, same run-derived shape), plus two signals:

| code | severity |
|---|---|
| `source_matching_zero` | warning |
| `source_matching_degraded` | info |

**No new collection.** It reads `extra_metrics.groups` on the newest completed `full_epg` run — the per-source breakdown #645 moved onto the parent row when per-group sub-runs were retired. Runs without it (in progress, or pre-v90) yield no opinion rather than a guess from the retired `event_group` rows.

## Threshold work — this is where the real design was

I measured against a live 447-source install rather than guessing, and the naive version would have been useless:

- **Denominator is `matched + unmatched`, not `fetched`.** `fetched` is pre-filter. "USA | NFL Teams Backup" fetches 34 streams in August and offers none to the matcher — at `fetched` it looks broken, and every out-of-season source would fire. Zero-denominator sources are absent from the result, not reported at 0%.
- **Minimum 10 post-filter streams.** 131 sources matched nothing, but the distribution is brutal:

  ```
  den >= 1 : 131      den >= 5  : 59
  den >= 2 : 104      den >= 10 : 41
  den >= 3 :  84      den >= 20 : 20
  ```

  72 of the 131 had fewer than five streams. One unmatched stream is not evidence.
- **Warning/info, not critical, and the evidence is capped and ranked.** Even at ≥10, ~41 sources legitimately match nothing — "US | News", "USA | Peacock LIVE TV", generic channel groups with no fixtures to find. Firing critical on 41 sources is a list, not an alarm. Sorted by stream count with a `total`, so truncation keeps the biggest offenders.

## Verified against that install’s real run

```
source_matching_zero      total=41
   200 streams  matched=0   [375] AU | Bar TV
   100 streams  matched=0   [76]  US| BTN+ PPV
   100 streams  matched=0   [97]  US| BTN+ PPV
   100 streams  matched=0   [451] USA: BTN+ [PPV EVENTS] [1080p]
    90 streams  matched=0   [268] Sports | Big Ten+

source_matching_degraded  total=50
    46%  1251 streams  [124] Dispatcharr Channels
    25%   486 streams  [95]  US| ESPN PLUS
    25%   486 streams  [465] USA: ESPN PLUS [1080p]
```

Four BTN+ sources at 0/100 and ESPN PLUS at 25% of 486 are real problems the bundle was silent about.

## Contract

Per the Support Bundle Contract in `CLAUDE.md`: implementation, tests (`tests/test_support_bundle.py`, +3 cases incl. the off-season-silence and missing-breakdown guards), and user docs (`docs/guide/troubleshooting.md`) all updated here. The bundled `AGENTS.md` renders signals dynamically so it picks these up with no change — asserted in the test. `SCHEMA_VERSION` stays 1: new codes are additive within the existing `signals` shape, and `evidence` is already per-code.

Gates: `2561 passed`, `ruff` clean, frontend builds.


egyptiangio (approved):
Verified end to end: `generation.py` writes `extra_metrics.groups` from `to_run_summary()` with the `matched`/`unmatched` keys this reads, `_row_to_run` JSON-decodes it, and evidence exposes only source id/name (already in the bundle). Contract check: impl + tests + `troubleshooting.md` updated, `AGENTS.md` renders dynamically, schema stays additive at v1. Bundle tests and ruff pass on the branch.

One wording nit, take it or leave it: `source_matching_health`'s docstring says "newest completed run" but the loop takes the newest `full_epg` run carrying a breakdown regardless of status. Behaviour is fine (a run that failed after matching still has real matching data) — just the docstring.

PR #659, opened 2026-08-29
fix(matching): trim pipe metadata from team names, stop apostrophes splitting them (#652, #653)

### LINKED ISSUE #652: Pipe-delimited metadata leaks into team2 — neither classifier nor matcher resolves it

lpukatch:
Closes #652 and #653. Two defects in the same shared normalization path, which is why they ship together — both were found auditing NCAAF on 2026-08-29, where between them they accounted for **all 22 failures** of one source.

## #652 — pipe metadata rides into the team name

Each side of the code believed the other handled this.

`classifier.py` deliberately preserves pipe content:

> preserve pipe content for fuzzy matching disambiguation.
> **The matcher will try both sides of the pipe and pick the one that matches.**

`team_matcher.py` asserted the opposite:

> is handled naturally by token_set_ratio which finds best token overlap.
> **No explicit pipe resolution needed.**

Neither was true. There was no pipe code anywhere in `team_matcher.py`, and `token_set_ratio` charges for the extra tokens rather than ignoring them:

```
"WAGNER | 8.29 | NEC FRONT ROW"          vs "Wagner Seahawks"   =  57.1
"WAGNER"                                 vs "Wagner Seahawks"   = 100.0
"TCU (IN DUBLIN, IRELAND) | 8.29 | ESPN" vs "TCU Horned Frogs"  =  38.3
```

with `BOTH_TEAMS_THRESHOLD` at 60. The streams that *did* match this shape were surviving on luck — `UNLV | 8.29 | FOX` only via the abbreviation path, `VIRGINIA | 8.29 | ESPN` clearing the floor by 1.5 points.

`_prepare_pipe_fallback` mirrors the existing parenthetical fallback and is wired into both candidate loops as the **last** tier, so a name that matches intact never reaches it. Only the **leading** segment is taken: by that point the classifier has already moved prefix noise (league hint, sport, provider, bare datetime) to the back, so what precedes the first pipe is the team and what follows is metadata. Trying every segment would let a venue or network name score as a team. Both stale comments are corrected.

### Dot-separated dates — proposed and withdrawn

A `MM.DD` pattern was in the first revision. Removed in review: UK listings write kickoff times the same way, so `Arsenal v Chelsea 7.30` read as 30 July. Built-in masks are `extracted_date_trusted=True` and the trusted-date gate hard-rejects candidates more than a day out, so it would have turned correct matches into `DATE_MISMATCH`. The two-digit-day rule was the worst possible filter, since `.30` and `.15` are the common minute values.

It was validated against a corpus of real failed matches that is US-feed dominated and contained no bare kickoff times — the wrong sample for a question about naming conventions.

## #653 — apostrophes split team names

The two normalizers disagree, and the matcher runs **both in sequence** — `_match_against_events` calls `normalize_for_matching`, then `_score_teams_against_event` calls `normalize_text` on the result — so the split is unrecoverable:

```python
normalize_for_matching("Hawai'i")  ->  'hawai i'   # [^\w\s] -> " "
normalize_text("Hawai'i")          ->  'hawaii'    # apostrophe deleted
```

```
'hawai i' vs 'hawaii rainbow warriors' =  46.7   <- what the matcher compared
'hawaii'  vs 'hawaii rainbow warriors' = 100.0
```

46.7 is under the floor, so the event was rejected despite a perfectly parsed stream — that matchup scored **100.0** on `_score_teams_against_event` called directly, yet returned `no_event_found` end to end. `normalize_for_matching` now deletes apostrophes and backticks before the punctuation rule, exactly as `normalize_text` does. unidecode has already run by then, so a curly apostrophe is covered.

## Measured

Against the 22 real failures of source 254 (verbatim from `epg_failed_matches`), replayed over the live ESPN slate:

**0/22 → 17/22.**

The other 5 are **not** pipe-related — verified by re-running them with no pipe metadata at all, where they still fail. They are short codes (`CCSU`, `UAPB`) and school-name variants (`GRAMBLING STATE` vs ESPN's `Grambling Tigers`, `WINSTON-SALEM STATE` vs `Winston-Salem Rams`) — a separate class, not filed as part of this.

## Known gap: the fallback tiers do not compose

One stream carries **both** noise classes and so needs both removed:

```
raw team2                 'TCU (IN DUBLIN, IRELAND) | 8.29  | ESPN'
paren-stripped  (tier 3)  'tcu 8 29'              -> 54.5
pipe-trimmed    (tier 4)  'tcu in dublin ireland' -> 43.2
pipe + paren    (missing) 'tcu'                   -> 100.0
```

Each tier removes one class, neither alone clears 60. The withdrawn dot-date pattern was accidentally propping this case up by emptying the pipe tail so tier 3 produced `tcu espn`, with ESPN then dropped as a broadcast network.

Composing the tiers is a small, well-motivated follow-up and would take this to 18/22. Left out of this PR rather than added after review — happy to fold it in here if preferred.

## Tests

`tests/matching/test_pipe_and_apostrophe.py`, 20 cases across three groups: both normalizers agree on apostrophes (incl. curly), the leading-segment rule and its 3-char floor, the documented `Sacramento Kings | Golden 1 Center` pass-through still resolving to the team, and the dot-date bounds — decimals, out-of-range months, dot-times, and am/pm all asserted **not** to parse as dates, with slash dates still winning.

One test deliberately asserts the classifier still hands over the pipe tail, so the fallback cannot be silently bypassed by a future classifier change.

Gates: `2588 passed`, `ruff` clean, frontend builds.

### DIFF
--- a/teamarr/consumers/matching/classifier.py
+++ b/teamarr/consumers/matching/classifier.py
@@ -1022,7 +1022,9 @@ def _clean_team_name(name: str) -> str:
     name = re.sub(round_pattern, "", name, flags=re.IGNORECASE | re.VERBOSE)
 
     # Handle "|" separator - preserve pipe content for fuzzy matching disambiguation
-    # The matcher will try both sides of the pipe and pick the one that matches.
+    # The matcher trims to the leading segment as its last fallback tier
+    # (TeamMatcher._prepare_pipe_fallback, #652) — that code did not exist when
+    # this comment was written, so the tail simply rode into the team name.
     # Here we only strip OBVIOUS prefix noise (league hints, channel numbers) from the
     # start, keeping the rest intact for the matcher to disambiguate.
     # "NFL | Bills vs Broncos" → "Bills vs Broncos" (NFL is league hint)
--- a/teamarr/consumers/matching/normalizer.py
+++ b/teamarr/consumers/matching/normalizer.py
@@ -273,19 +273,6 @@ DATE_PATTERNS = [
     (rf"\b(\d{{1,2}})(?:st|nd|rd|th)?\s+(?:{_MONTH_NAMES})\b", "DATE_MASK"),
     # Dec 31, December 31 - use negative lookahead (?!:) to avoid matching "Jan 11:45pm"
     (rf"\b(?:{_MONTH_NAMES})\s+(\d{{1,2}})(?:st|nd|rd|th)?(?!:)\b", "DATE_MASK"),
-    # 8.29, 08.29 (M.DD without year) — dot-separated dates (#652). Deliberately
-    # LAST: masking happens whether or not the date parses, so a greedy pattern
-    # here would eat text the forms above read correctly, and the loop breaks on
-    # first match.
-    #
-    # Tightly bounded, because a dot between digits is usually NOT a date.
-    # Measured over 1036 dot-numbers in real stream names: 1020 had a two-digit
-    # day and every one was a date (08.27, 8.29); the one-digit-day cases were
-    # noise ("5.1"), and "17.45"/"10.30" were TIMES written with a dot. So the
-    # month is bounded 1-12 (killing 13.4 and 17.45), the day must be two digits
-    # 01-31 (killing 5.1 and decimal spreads like "2.5"), and an am/pm suffix
-    # vetoes the read so "10.30pm" stays a time.
-    (r"\b(1[0-2]|0?[1-9])\.(3[01]|[12]\d|0[1-9])\b(?!\s*[ap]\.?m)", "DATE_MASK_NO_YEAR"),
 ]
 
 # Time patterns to extract and mask (with optional TZ suffix)
--- a/teamarr/consumers/matching/normalizer.py
+++ b/teamarr/consumers/matching/normalizer.py
@@ -273,6 +273,19 @@ DATE_PATTERNS = [
     (rf"\b(\d{{1,2}})(?:st|nd|rd|th)?\s+(?:{_MONTH_NAMES})\b", "DATE_MASK"),
     # Dec 31, December 31 - use negative lookahead (?!:) to avoid matching "Jan 11:45pm"
     (rf"\b(?:{_MONTH_NAMES})\s+(\d{{1,2}})(?:st|nd|rd|th)?(?!:)\b", "DATE_MASK"),
+    # 8.29, 08.29 (M.DD without year) — dot-separated dates (#652). Deliberately
+    # LAST: masking happens whether or not the date parses, so a greedy pattern
+    # here would eat text the forms above read correctly, and the loop breaks on
+    # first match.
+    #
+    # Tightly bounded, because a dot between digits is usually NOT a date.
+    # Measured over 1036 dot-numbers in real stream names: 1020 had a two-digit
+    # day and every one was a date (08.27, 8.29); the one-digit-day cases were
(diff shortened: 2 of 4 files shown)


egyptiangio (changes requested):
Apostrophe fix and pipe fallback look right — thanks. One blocker: please drop the `DATE_PATTERNS` dot-date entry and `TestDotSeparatedDates`.

Verified on this branch:

```
Arsenal v Chelsea 7.30              -> 2026-07-30
Sky Sports: Man Utd v Spurs 10.30   -> 2026-10-30
Liverpool vs Everton 12.30 KO       -> 2026-12-30
```

Built-in masks are `extracted_date_trusted=True` and `team_matcher.py:1225` hard-rejects >1-day mismatches, so every UK bare kickoff time becomes a `DATE_MISMATCH` on a correct match. The am/pm veto only catches `10.30pm`.

It's also not load-bearing for #652 — with `8/29` fully extracted, team2 still scores 57.1; the pipe fallback alone fixes it. Full reasoning in the #652 comment. Happy to merge as soon as that's out.

egyptiangio (approved):
Dot-date pattern dropped — thanks. Apostrophe rule and pipe fallback verified earlier; merging.

Issue #662, opened 2026-08-29
74% of match failures land in catch-all buckets; 'unmatched' is a string fallback, not a FailedReason


lpukatch:
Found auditing NCAAF on 2026-08-29, alongside #661.

## 38% of failures carry no reason at all

Live install, 13:00–15:00 window:

```
unmatched              4279   <- not a FailedReason; a string fallback
no_event_found         4050
fixture_not_in_league  1638
team2_not_found         614
no_tennis_match         356
no_event_card_match     198
team1_not_found          90
date_mismatch            54
```

`unmatched` is not a member of `FailedReason`. It is the literal in `consumers/event_group_processor/persistence.py`:

```python
failed_reason = "unmatched"
if result.failed_reason:
    failed_reason = result.failed_reason.value
```

So 4,279 results reached persistence with `failed_reason` unset. Combined with `no_event_found`, **74% of all failures land in a bucket that does not say why.**

## Why it matters

`no_event_found` is documented as "Teams matched, league detected, no game scheduled" — a specific claim. In practice it also absorbs cases where the teams scored 100/100 and something else rejected the candidate, which is how #650 stayed invisible.

## Suggested

1. Find the paths that return a failed outcome without setting `failed_reason` and give them one. If any are genuinely unclassifiable, an explicit `UNCLASSIFIED` member is still better than a bare string, because it is greppable and shows up in the taxonomy honestly.
2. Split `no_event_found` so "no candidate scored high enough" is distinguishable from "candidates existed and were rejected by a gate". The candidate loop already counts `fixture_rejected` and `date_rejected` — those counters currently inform nothing that is persisted.

Best done after (or with) #661, since the detail field is where the supporting evidence belongs.

lpukatch:
Bead: `teamarrv2-oteh`

egyptiangio:
Bead: `teamarrv2-xu7x` (blocked by `lux2` / #661)

egyptiangio:
The numbers are right but the cause isn't — building from the stated cause would fix the wrong thing.

[...]

Issue #660, opened 2026-08-29
Candidate-matching loop is duplicated across single/multi-league paths (91%) and has already drifted


lpukatch:
Found auditing NCAAF on 2026-08-29. This is not a hypothetical maintenance concern — the drift it produced is the root cause of #650.

## The duplication

| function | lines |
|---|---|
| `TeamMatcher._match_against_events` | 245 |
| `TeamMatcher._match_against_multi_league_events` | 256 |

**91% of the smaller function is line-for-line identical to the larger.**

## It has already diverged, and the divergence caused a P0

Strip the plumbing differences (tuple-vs-plain events, `best_league` tracking, log format) and one behavioural difference remains — the fixture gate:

```python
# _match_against_multi_league_events        # _match_against_events
if (self._fixture_vetoes(...)               if self._fixture_vetoes(...):
    and event.league not in                     fixture_rejected += 1
        hinted_leagues):                        continue
```

That escape hatch arrived in `b4fae42e` (2026-08-27), *"fix(matching): honor explicit league hints in fixture gate (#627)"*, whose own comment reads:

> An explicit league hint already narrowed this stream to the candidate's league. It is stronger evidence than an **incomplete cross-league identity index**; unhinted streams retain the gate.

"Incomplete cross-league identity index" is a verbatim description of #650's root cause. And the affected streams all carry the hint:

```python
classify_stream('NCAAF 17: Fairmont State at Dayton 12pm').league_hint
# -> 'college-football'
```

So the mitigation for #650's failure mode existed **two days before #650 was diagnosed** — in one of the two copies. The single-league path, which those sources use, kept the bug and produced 512 `fixture_not_in_league` failures in one support bundle.

## It is still actively dangerous

The #652 pipe fallback required the *identical* edit in both loops. Missing one would have silently split behaviour by source type — single-league sources fixed, multi-league sources not — with no test catching it, because no test asserts the two paths agree.

## Suggested direction

Extract the shared per-candidate body into one function taking `(ctx, league, event, ...)` and returning a scored candidate or None, leaving each caller to own only what genuinely differs (event iteration shape, `best_league` tracking, result construction).

Whatever the approach, a test that asserts the two paths reach the same verdict for the same stream+event is the thing that would have caught #627's half-landing.

lpukatch:
Bead: `teamarrv2-z4bs`

egyptiangio:
Bead: `teamarrv2-7n3a`

[...]

PR #663, opened 2026-08-29
fix(matching): persist the near-miss diagnosis on failed matches (#661)

### LINKED ISSUE #661: Match failures persist no detail — the near-miss diagnosis is computed, logged, then truncated away
Found auditing NCAAF on 2026-08-29. Diagnosing four matching bugs took a full manual investigation — reading 1026 JSON records and re-deriving causes from `parsed_team1`/`parsed_team2` string shapes — because the structured failure records carry no explanation. The process had already computed the explanation and thrown it away.

## Every failure record has a null detail

Measured on a live install, 13:00–15:00 window:

```
failures = 11279
with NO detail = 11279  (100%)
```

The plumbing is entirely in place:

- `MatchOutcome.failed(...)` accepts and stores `detail`
- `FailedMatch` declares `detail: str | None = None`
- `epg_failed_matches.detail` exists in the schema
- the support bundle ships the column

`consumers/event_group_processor/persistence.py:155` constructs `FailedMatch(...)` without passing `detail`. That single omission empties the column.

## The diagnosis is computed, then written somewhere it cannot survive

`TeamMatcher._log_near_miss` already produces exactly what a triager needs — the closest candidate, per-side scores against the threshold, alias resolution, and whether the date gate fired. It emits to a DEBUG log.

From the reporter's own bundle:

```
[NEAR_MISS] stream_id=280906 best='South Dakota Coyotes vs Central Connecticut Blue Devils'
  (college-football 2026-08-29) scores Central Connecticut=100 / South Dakota=100 (need 60)
  alias1=none alias2=none date_gated=0
```

**Both sides scored 100 against a threshold of 60, and the match still failed.** That is an unmissable signature of a non-scoring rejection (the fixture gate, #650). It sat in the log while the structured record said `no_event_found` with no detail.

A log is also the wrong home for it. The bundle caps log tails at `LOG_BYTES = 256 KB`, so **78** near-miss lines survived out of 11,279 failures — 0.7%. The structured table, which ships complete at 500 rows per run, had the empty column.

## Suggested

1. Pass the `detail` that `MatchOutcome` already carries through `persistence.py`.
2. Persist the near-miss summary as structured detail on the failure record rather than only logging it — same content `_log_near_miss` builds today, capped in length.
3. Since it would no longer depend on DEBUG being enabled, the data becomes available on every install rather than only where file logging is at DEBUG.

Pairs with #662: once the reason is always set and the detail always populated, the catch-all buckets stop hiding distinct causes.

lpukatch:
Closes #661. First of the three structural findings from the 2026-08-29 matching audit (#660, #661, #662).

## Problem

Every failure record shipped with `detail` NULL — **11,279 of 11,279** in a real support bundle. Diagnosing four matching bugs that day meant reading 1026 JSON records and re-deriving causes from `parsed_team1`/`parsed_team2` string shapes.

The explanation already existed. `_log_near_miss` computed the closest candidate, both per-side scores against the threshold, alias resolution and date gating — then wrote it to a DEBUG log, where two things destroyed it:

1. It early-returned unless DEBUG was enabled, so installs that raise `LOG_LEVEL` never computed it at all.
2. The bundle caps log tails at `LOG_BYTES = 256 KB`, so **78** near-miss lines survived out of 11,279 failures — 0.7%.

Meanwhile the structured table the bundle ships *complete* (500 rows per run) had the empty column.

The surviving lines show what was being thrown away:

```
[NEAR_MISS] stream_id=280906 best='South Dakota Coyotes vs Central Connecticut Blue Devils'
  scores Central Connecticut=100 / South Dakota=100 (need 60) alias1=none alias2=none date_gated=0
```

Both sides scored 100 against a threshold of 60 and it still failed — an unmissable signature of a non-scoring rejection (the fixture gate, #650). The structured record said `no_event_found`, no detail.

## Change

- `_log_near_miss` becomes `_near_miss_summary`, **returning** the line instead of logging it, and no longer gated on DEBUG. The caller still logs it, and now also hangs it on the `MatchOutcome`.
- `persistence.py` passes the `detail` it had always been dropping. `MatchOutcome.failed()`, the `FailedMatch` dataclass, the schema column and the INSERT **all already supported it** — only the constructor call omitted it. So this also recovers details other paths were already setting, e.g. `"No events in college-football for 2026-08-29"`.
- Capped at `NEAR_MISS_DETAIL_MAX = 400`, informative half first, since it rides in every bundle at up to 500 rows per run.

Wired at both failure sites — the single-league and multi-league loops, which are 91% duplicated (#660).

## Result

The NCAAF failure that took an afternoon to diagnose now reads:

```
best='TCU Horned Frogs vs North Carolina Tar Heels' (college-football 2026-08-29)
scores ROBERT MORRIS=41 / WAGNER | 8.29  | NEC FRONT ROW=41 (need 60)
alias1=none alias2=none date_gated=0
```

That names the #652 pipe bug on sight — the polluted team string is right there with its score.

## Cost

Bounded at 300 candidates and only on the failure path; the per-side scores come from the same memoized kernel the match loop just used, so they are overwhelmingly cache hits. In practice this is not new work on most installs: file logging already defaults to DEBUG, so the computation was already running — it just had nowhere durable to go.

## Redaction

No change needed, and no new exposure. The bundle's `_sanitize` recurses into run details and strips URLs, tokens and M3U account names from every string, so `detail` is covered exactly as `stream_name` and `parsed_team1/2` already are — and its content is drawn from those same fields.

## Tests

`tests/matching/test_failure_detail.py`, 7 cases: detail populated on failure, names the closest candidate and both scores against the threshold, the empty-window path explains itself, the length cap holds, the summary no longer depends on DEBUG being enabled, and the value survives a round trip through `save_failed_matches`. One test asserts the exact `persistence.py` constructor line, since a silently-omitted kwarg is what caused this.

Gates: `2575 passed`, `ruff` clean, frontend builds.

Note: `tests/test_cache_eviction.py::test_insert_into_a_roomy_cache_does_not_scan` failed once during this work and passes in isolation and on 3/3 subsequent full runs. It is a wall-clock timing assertion by design ("Timed rather than mocked because the defect was purely asymptotic"), so it is load-sensitive — unrelated to this change, but worth knowing it can flake in CI.

### DIFF
--- a/teamarr/consumers/event_group_processor/persistence.py
+++ b/teamarr/consumers/event_group_processor/persistence.py
@@ -161,6 +161,16 @@ class MatchPersistence:
                         stream_id=stream_id,
                         stream_name=result.stream_name,
                         reason=failed_reason,
+                        # The evidence behind the reason (#661), and the
+                        # inclusion verdict beside it (#662's unmatched rows).
+                        # MatchOutcome has carried a detail since it was
+                        # introduced and both columns have always existed, but
+                        # this constructor passed neither, so every one of
+                        # 11,279 failures in a real bundle shipped with detail
+                        # NULL and triage had to re-derive causes from the
+                        # parsed_team strings.
+                        detail=result.detail,
+                        exclusion_reason=result.exclusion_reason,
                         parsed_team1=parsed_team1,
                         parsed_team2=parsed_team2,
                         detected_league=detected_league,
--- a/teamarr/consumers/matching/constants.py
+++ b/teamarr/consumers/matching/constants.py
@@ -9,6 +9,12 @@ For pattern/alias data, see teamarr/utilities/constants.py
 # The lifecycle layer filters out past events after matching.
 MATCH_WINDOW_DAYS = 30
 
+# Longest near-miss explanation persisted on a failed match (#661). It rides in
+# every support bundle at up to 500 failure rows per run, so it is capped to
+# keep the archive bounded; the informative half (best candidate + both scores)
+# comes first, so a truncation loses only the alias/date tail.
+NEAR_MISS_DETAIL_MAX = 400
+
 # =============================================================================
 # CONFIDENCE THRESHOLDS
 # Fuzzy match score thresholds that control when matches are accepted.
--- a/teamarr/consumers/matching/team_matcher.py
+++ b/teamarr/consumers/matching/team_matcher.py
@@ -25,6 +25,7 @@ from teamarr.consumers.matching.constants import (
     ALTERNATE_TEAM_CODES,
     BOTH_TEAMS_THRESHOLD,
     HIGH_CONFIDENCE_THRESHOLD,
+    NEAR_MISS_DETAIL_MAX,
     SHORT_CODE_MAX_LEN,
 )
 from teamarr.consumers.matching.country_resolver import (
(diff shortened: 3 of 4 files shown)


egyptiangio (commented):
Reviewed — the change is right and small. One request: #664 (#660) just collapsed the two candidate loops into `_match_against_candidates`, so after it landed this needs a rebase and your edit becomes a single site (the shared tail).

Two nits, take or leave: `persistence.py` also drops `exclusion_reason` the same way — worth the second one-liner while you're there (it's what #662's `unmatched` rows need); and `getattr(result, "detail", None)` can be plain `result.detail`, it's a typed field.

Will approve once rebased.

lpukatch:
Rebased onto dev (`cae83279`) — all three points addressed.

**Rebase.** As predicted, the conflict was the whole old single-league copy, so I reset `team_matcher.py` to dev's version and re-applied onto `_match_against_candidates`. The diff there is now one site: `detail=near_miss` on the shared tail's `MatchOutcome.failed`, plus the `_log_near_miss` → `_near_miss_summary` conversion.

**`exclusion_reason`.** Confirmed the same drop — the field exists on `FailedMatch` (`stats.py:669`), the INSERT carries it (`stats.py:734`), and `_outcome_to_result` sets it for every failed outcome (`matcher.py:1701`); the unmatched branch just never passed it. Added next to `detail`, with the reasoning folded into the existing comment rather than a second block.

**`getattr`.** Now plain `result.detail` — you're right, `detail: str | None = None` is typed on `MatchedStreamResult` (`matcher.py:156`) and always set, so the `getattr` guarded nothing.

One thing beyond the review: `test_processor_passes_the_outcome_detail_through` asserted the literal source string `detail=getattr(result, "detail", None)` via `inspect.getsource`, so it would have failed on nit 2 by construction — it was pinning the typo, not the behavior. Replaced with a test that drives `_save_match_details` against a real connection and reads both columns back off the persisted row.

Gates: ruff clean, 2596 passed, frontend build green.

egyptiangio (approved):
All three addressed, single site on the shared tail, and the round-trip test is the right replacement for the source-string assertion. Merging.

Issue #717, opened 2026-09-05
Add {division} channel-group wildcard (FBS/FCS)


lpukatch:
## Problem

A college-football Saturday is 70-100 games. With a static or `{league}` channel group they all land in one Dispatcharr group, which is unparseable. `{conference}` (#91) helps but fans it out to 25 groups for CFB, ~14 of which are FCS.

## Proposal

Add a `{division}` wildcard to dynamic channel-group patterns, resolving to the ESPN parent group of the home team's conference:

- `college-football` → **FBS** / **FCS**
- `mens-college-basketball` / `womens-college-basketball` → **Division I**

`{league} | {division}` gives "NCAAF | FBS" and "NCAAF | FCS" — two groups instead of one bucket of 100 or 25 conference groups.

## Why this is cheap

`CacheRefresher.CONFERENCE_LEAGUES` already walks roots `("80", "81")` = FBS/FCS and discards the parent. ESPN's group payload carries `parent.$ref` and the root group's own `name`/`shortName` are literally "FBS"/"FCS" — no maintained mapping, no new provider integration.

Deliberately **not** doing Power 4 / Group of 5: ESPN exposes no tier flag for football (checked group payloads for both the conference and root levels), so it would require a hand-maintained conference→tier list that realignment churns every year.

## Cross-division games

`{conference}` buckets on the **home** team, and `{division}` inherits that. In FBS-vs-FCS games the FCS side is essentially always the paid road team, so those land in FBS — which is where you'd want them.

## Scope

- `provider_group_cache`: add `parent_key` / `parent_name` (column addition; schema reconciliation handles the upgrade)
- `_fetch_conference_tree`: read the root group's meta once per root, tag children with it
- `get_team_group`: return the division alongside name/abbrev
- `DynamicResolver`: `get_event_division` + `{division}` in `resolve_pattern` / `resolve_channel_group`, with the same static-group fallback as `{conference}` when it can't resolve
- Frontend pattern hints + `docs/guide/channels/output.md`

lpukatch:
Bead: `teamarrv2-yuo7`

lpukatch:
Landed on dev in 2d5a0b9b (PR #718). Closes at the next release.

egyptiangio:
Released in [v2.16.0](https://github.com/Pharaoh-Labs/teamarr/releases/tag/v2.16.0).

PR #816, opened 2026-09-13
feat: {conference_abbrev} wildcard for channel group patterns

### LINKED ISSUE #777: [Feature]: Option for abbreviations in the {conference} placeholder for channel group naming
### Problem / Use case

For the {conference} placeholder when using custom patterns in channel group naming, I would like an option to include them as abbreviations. For instance, Atlantic Coast Conference would become ACC. Some of the channel group names get too long (for me) when it writes out the full conference name.

### Proposed solution

Add a option like {conference_abbrev}, that would display an abbreviated version of the NCAA conference names. Similar to the way it works for team names in templates.

### Alternatives considered

_No response_

### Affected area

Custom patterns in channel group naming

### Additional context

_No response_

lpukatch:
Closes #777.

## Problem
Custom channel-group patterns with `{conference}` write out the full NCAA conference name ("Atlantic Coast Conference"), making group names longer than some users want.

## Solution
Add a `{conference_abbrev}` wildcard that resolves to the conference's compact form ("ACC", "SEC", "Big Ten"). The abbreviation data already existed — `provider_group_cache.group_abbrev` has been cached from ESPN's `shortName` since the `{conference}` wildcard landed (#91), and `get_team_group` already returned it; the resolver just never exposed it.

Implementation mirrors `{division}` (#717) exactly:
- `DynamicResolver.get_event_conference_abbrev()` — from the home team's cached conference (shares the same per-team lookup cache as `{conference}`/`{division}`)
- `resolve_pattern()` — new `conference_abbrev` param + exact-token replace (cannot collide with `{conference}` since the brace follows `_`)
- `resolve_channel_group()` — resolves it for pattern mode; an event with no abbreviation (non-NCAA, or a tree row without one) falls back to the static group, same contract as the other wildcards

## Example
`{league} | {conference_abbrev}` → "NCAAF | SEC" instead of "NCAAF | Southeastern Conference"

## Changes
- `teamarr/consumers/lifecycle/dynamic_resolver.py` — the wildcard
- `DispatcharrOutputSettings.tsx`, `PerLeagueChannelConfig.tsx` — pattern help text
- `tests/test_provider_groups.py` — 5 new tests (resolution, coexistence with `{conference}`, home-team lookup, missing-abbrev deferral, static-group fallback)
- `docs/guide/channels/output.md`, `docs/guide/dispatcharr-integration.md`, `CLAUDE.md` — wildcard documentation

## Testing
- `pytest tests/` — 3274 passed
- `npm run build` — clean

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -286,7 +286,7 @@ All `update_channel` calls go through `_safe_update_channel`, which checks `Oper
 - API routes build responses with `to_model(Model, dataclass)` from `api/routes/settings/models.py`; frontend hooks are factory-generated with scoped cache invalidation (`frontend/src/hooks/useSettings.ts`).
 
 **Dynamic Groups** (`teamarr/consumers/lifecycle/dynamic_resolver.py`):
-- `{sport}`, `{league}`, and `{conference}` wildcards (`{conference}` = home team's NCAA conference from `provider_group_cache`, #91)
+- `{sport}`, `{league}`, `{conference}`, `{conference_abbrev}`, and `{division}` wildcards — the conference ones resolve the home team's NCAA conference from `provider_group_cache` (#91; abbrev #777, division #717)
 - Auto-creates in Dispatcharr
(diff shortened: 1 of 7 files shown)


lpukatch:
Follow-up commit: the profile-pattern path only checked `{sport}`/`{league}` for unresolved wildcards, so a *profile* pattern like `{league} | {conference}` would have created a Dispatcharr profile literally named `NCAAF | {conference}`. Now guarded for all conference tokens — profiles resolve from sport/league only.

PR #843, opened 2026-09-15
Replace fixed-cron EPG scheduling with pre-match + discovery triggers

Nathant93:
The EPG generation scheduler no longer runs on a fixed cron expression. Instead it wakes up either:

- shortly before the earliest known upcoming match (pre_match_lead_minutes, default 30), refreshing lineups/venue/broadcast info right before kickoff, or
- on a periodic discovery fallback (epg_discovery_interval_hours, default 4) that runs generation even with nothing scheduled soon, so newly added teams/leagues and far-future matches still get found.

Whichever trigger is sooner wins, and both are recalculated after every run. A slate of matches starting close together only needs one pre-match run, not one per match, since a single run refreshes the whole EPG.

Backup and scheduled channel reset are unaffected - they remain on their own true cron schedules via SubTaskScheduler, since they're time-of-day driven rather than match-driven.

- teamarr/consumers/scheduler.py: CronScheduler now computes next_run from the two triggers instead of croniter; added _get_next_match_start, _compute_next_run, _mark_pre_match_covered, _parse_event_dt.
- database/schema.sql, database/settings/types.py: added pre_match_lead_minutes and epg_discovery_interval_hours to EPGSettings; cron_expression kept as a deprecated, unused column for backward compat.
- services/scheduler_service.py, api/routes/settings/{models,epg,lifecycle}.py, api/app.py: updated SchedulerStatus and settings endpoints accordingly.
- frontend: GeneralTab.tsx replaced the cron expression input with lead time / discovery interval controls and a next-run reason indicator.
- tests/test_pre_match_scheduler.py: new coverage for the trigger logic; fixed tests/subscriptions/test_sub_scheduler.py call sites that used the old cron_expression constructor kwarg.

Also changed: the scheduler now runs once on startup (previously it did not), since the dynamic schedule has nothing to anchor to until a generation run has populated managed_channels with upcoming match times.


egyptiangio:
Thanks for putting real work into this, the pre-match trigger is a genuinely useful idea. Two mechanics before it can be reviewed: please retarget to `dev` (main only receives releases) and open an issue describing the design so we can discuss it there.

On the substance: cron scheduling has to stay. Beyond the many installs that rely on a predictable schedule, regular generation matters *after* events as much as before them. Final scores, results and post-game state only reach the EPG because generation keeps running at intervals once the games are over, and a schedule anchored to upcoming kickoffs never covers that. So this fits best as an additional scheduling mode that layers a pre-match wake-up on top of the cron schedule, rather than replacing it. Same for the startup run: opt-in rather than always-on.

Let's settle that shape on the issue first, then I'm happy to review the PR against it.

Nathant93:
Thanks I have opened a new PR - I will put in a "issue" as well

PR #849, opened 2026-09-16
Event-driven EPG scheduling (with cron kept as an option)

Nathant93:
### Summary

Replaces the fixed-cron trigger for EPG generation with an event-driven scheduler that runs generation ahead of each match's actual start time, instead of on an arbitrary fixed interval. The classic cron schedule is kept as a selectable mode for anyone who prefers the old, more predictable behavior.

### What changed

**Scheduling modes** (`Settings → General → Schedule`)
- **Before each match** *(new default)* — two triggers, whichever fires first:
  - **Pre-match**: runs generation `pre_match_lead_minutes` (default 30) before the earliest known upcoming match, so lineups/venue/broadcast info stay fresh right up to kickoff instead of going stale between cron ticks.
  - **Discovery**: a fallback every `epg_discovery_interval_hours` (default 4) that runs generation even with nothing scheduled soon, so newly added teams/leagues and far-future matches still get found.
  - A whole slate of matches starting close together (e.g. a Sunday 1pm NFL slate) is covered by one run, not one per match.
- **Fixed cron schedule** — unchanged classic behavior, croniter-driven, kept for anyone who wants predictable fixed-time runs instead.

**Per-sport lead times**
- New `sport_schedule_overrides` table lets individual sports override the global pre-match lead time (e.g. 90 min for NFL pregame coverage vs. 15 min for something faster-moving).
- Matches are ranked by their own effective trigger time (start minus that sport's lead), not raw start time — a sport with a longer lead can fire before an earlier-starting match with a shorter one.
- Overrides are read fresh from the DB on every scheduler computation, so changes apply on the next cycle with no restart required.

### Backend
- `teamarr/consumers/scheduler.py` — `CronScheduler` takes a `mode` param (`pre_match` | `cron`); pre-match logic reworked around `_get_next_trigger()`/`_MatchTrigger` to account for per-sport leads.
- `teamarr/database/sport_schedule.py` — CRUD for the overrides table.
- `schema.sql`, `database/settings/types.py` — new `scheduler_mode`, `pre_match_lead_minutes`, `epg_discovery_interval_hours` settings fields; `cron_expression` retained and active in cron mode.
- `api/routes/settings/{epg,lifecycle,models}.py` — mode/cron validation, new `/settings/scheduler/sport-lead-times` GET/PUT/DELETE endpoints.
- `services/scheduler_service.py`, `api/app.py` — mode/cron_expression threaded through `SchedulerStatus` and startup wiring.

### Frontend
- Scheduling-mode dropdown in `GeneralTab.tsx`, conditionally showing cron controls or lead-time/discovery controls.
- New `SportLeadTimeManager` card for per-sport overrides, shown only in pre-match mode.

### Behavior changes worth flagging
- The scheduler now runs once on container startup in pre-match mode (previously it never did), since the dynamic schedule needs at least one run to discover upcoming matches before it can schedule around them. Cron mode keeps the original no-run-on-start behavior.

### Testing
- `tests/test_pre_match_scheduler.py` — 26 tests covering trigger computation, per-sport overrides, and cron mode.
- Full backend suite: 3078 tests passing.
- Frontend: typechecks, lints, and builds clean.


justinglock40:
Awesome this would fix an issue I discovered myself today. Would this run a full generation each time or would it use the sync window to just update Dispatcharr?

Nathant93:
> Awesome this would fix an issue I discovered myself today. Would this run a full generation each time or would it use the sync window to just update Dispatcharr?

It's a full generation run as a targeted generation was a lot of effort

